PR #54, opened 2026-01-20
feat: Add disable_shift config option for Colima/Lima environments

technicalpickles:
## Summary

When running Incus inside a Colima/Lima VM on macOS, the `shift=true` option for disk devices fails with:

```
Error: Failed to start device "workspace": Required idmapping abilities not available
```

This happens because Lima already performs UID mapping at the VM level via virtiofs, and nested idmapping (Lima + Incus) causes conflicts.

This PR adds a `disable_shift` config option that allows users to disable UID shifting for these environments.

## Changes

- Add `DisableShift bool` field to `IncusConfig` struct
- Add `DisableShift bool` field to `SetupOptions` struct
- Integrate with existing `useShift` logic in session setup (line 163)
- Add informative log message: `"UID shifting disabled (Colima/Lima environment)"`
- Thread config value through shell.go and run.go

## Usage

```toml
# ~/.config/coi/config.toml
[incus]
disable_shift = true
```

## Technical Details

Lima mounts macOS directories into the VM via virtiofs and maps UIDs at the VM level:

```bash
$ mount | grep virtiofs
mount0 on /Users/josh type virtiofs (rw,relatime)

$ ls -la /Users/josh/myproject
drwxr-xr-x 25 lima lima 800 Jan 20 09:20 .
```

When Incus then tries to apply `shift=true` on top of this, the kernel capabilities for nested idmapping aren't available.

The workaround is to omit `shift=true` - Lima's VM-level mapping already handles file ownership correctly.

## Test plan

- [x] Build succeeds with `go build`
- [x] Config option is read correctly from `~/.config/coi/config.toml`
- [x] Log message appears: `[setup] UID shifting disabled (Colima/Lima environment)`
- [x] Workspace mounts successfully without `shift=true`
- [x] Container starts and runs normally

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-24.04
     timeout-minutes: 5
     steps:
-      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Set up Go
         uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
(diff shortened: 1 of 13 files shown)


mensfeld:
@technicalpickles i slightly updated the code so coi can auto-detect that its macos - can you please check if this is as expected? :pray:

mensfeld:
This PR has been superseded by #58 which is based on a branch in the main repository instead of a fork. This avoids the need for CI approval on each commit.

All changes from this PR have been merged into the new branch.

PR #329, opened 2026-04-14
Fix alias resolution: auto-register from CWD + prevent profile override

mensfeld:
## Summary

- **CWD alias auto-registration**: `ResolveAliasForLaunch()` now falls back to checking the CWD's `.coi/config.toml` for a matching alias when the registry lookup fails, fixing the chicken-and-egg problem where `coi shell <alias> --profile X` required a prior plain `coi shell` run
- **Profile alias no longer overrides project alias**: `applyContainerConfig()` only applies profile alias when the project has no alias set, preventing conflicts when the same profile is used across multiple projects
- Adds Go unit tests for `loadProjectAlias()`, CWD fallback logic, and alias preservation in `applyContainerConfig()`
- Adds Python integration tests for CWD auto-registration and profile alias override prevention

## Test plan

- [ ] `go test ./internal/alias/... ./internal/config/...` — all pass
- [ ] `coi shell karafka2 --profile rdkafka` from karafka2 dir with `[container] alias = "karafka2"` — works without prior registration
- [ ] `coi shell nonexistent` — still fails with "not found" error
- [ ] Profile with alias applied to project WITHOUT its own alias — profile alias still works
- [ ] Profile with alias applied to project WITH its own alias — project alias wins
- [ ] Two projects with own aliases using same profile (which has alias) — no conflict
- [ ] `pytest tests/cli/test_alias.py` — integration tests pass

### DIFF
--- a/internal/alias/resolve.go
+++ b/internal/alias/resolve.go
@@ -89,10 +89,17 @@ func ResolveAliasForLaunch(arg string) (*ResolvedAlias, error) {
 		cwd, cwdErr := os.Getwd()
 		if cwdErr == nil {
 			if projectAlias := loadProjectAlias(cwd); projectAlias == alias {
-				absWorkspace, _ := filepath.Abs(cwd)
+				absWorkspace, err := filepath.Abs(cwd)
+				if err != nil {
+					return nil, fmt.Errorf("failed to resolve workspace path for alias %q: %w", alias, err)
+				}
 				// Auto-register so subsequent lookups succeed.
-				_ = reg.Register(alias, absWorkspace, "")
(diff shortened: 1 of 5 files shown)


mensfeld:
The two CI failures are unrelated to this PR:

1. **Unit Tests** — `The operation was canceled` (CI infrastructure timeout, not a test failure)
2. **Integration Tests (monitoring-response)** — `test_file_read_at_threshold_triggers` flaky test in `test_security_monitoring.py` (`Container should be paused on 75MB read`) — timing-sensitive assertion, no relation to alias changes

Could you re-run the failed jobs?

PR #352, opened 2026-05-04
fix(cli): forward env vars to tmux windows beyond the first

SimonArnu:
## Summary

The previous implementation inlined `export KEY=VAL; ...` into the command string of the first pane only. That had two problems:

- The exports showed up in `ps auxww`, leaking secrets like `GITHUB_TOKEN`.
- Variables never reached the tmux server's update-environment, so any window or pane created later (e.g. `tmux new-window`, splitting a pane) inherited an empty environment.

This PR routes forwarded variables through tmux's native channels:

- `tmux new-session -e KEY=VAL ...` for the initial pane (not visible in `ps`).
- `tmux set-environment -g -t <session> KEY VAL` so subsequent windows/panes inherit them as well.
- Re-applied on attach so sessions created by older binaries are upgraded in place.

The two helpers (`buildTmuxNewSessionCmd`, `buildTmuxSetEnvironmentCmds`) are extracted so they can be unit-tested for ordering, quoting, and shell-injection-safe values.

## Test plan

- [x] `go test ./internal/cli/` passes — 6 new unit tests cover env-flag emission, deterministic ordering, awkward values (spaces, single quotes, `$HOME`), empty env, and `set-environment` command shape.
- [ ] Manual: open a coi session, run `tmux new-window` inside it, verify forwarded vars (`GITHUB_TOKEN`, `JIRA_*`) are present in the new window.
- [ ] Manual: confirm `ps auxww` no longer shows variable values.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -207,6 +207,15 @@ jobs:
           fi
           # Verify swap is active
           swapon --show
+          # Debug: show available cgroup controllers
+          echo "Root cgroup controllers:"
+          cat /sys/fs/cgroup/cgroup.controllers
+          echo "Root cgroup subtree_control:"
+          cat /sys/fs/cgroup/cgroup.subtree_control
+          # Enable memory controller in subtree if not already
+          echo "+memory" | sudo tee /sys/fs/cgroup/cgroup.subtree_control || true
+          # Check if memory.swap.max exists (indicates swap controller support)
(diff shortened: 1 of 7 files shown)


mensfeld:
Thank you. I will add integration tests for this, merge and release.

mensfeld:
Sorry for the noise, fixing someting alongside :pray:

mensfeld:
Ok this spec is not related so I will tackle it in another PR. Looks great.

mensfeld (approved):

PR #360, opened 2026-05-05
refactor(container): remove sg dependency, run incus directly

mensfeld:
## Summary

Follow-up to #358. Removes the `sg` (setgid) dependency entirely instead of conditionally using it.

- `sg` was used to activate the `incus-admin` group for each incus command, allowing COI to work immediately after `usermod -aG` without re-login
- In practice: adds complexity, breaks on distros where `sg` is restricted to root (ALT Linux), and users typically re-login after group changes anyway
- All platforms now run `incus` directly via `sh -c` — same path macOS always used

**Requirement:** Users must have `incus-admin` group membership active in their session (log out and back in after `usermod -aG`).

## Changes

- **`internal/container/commands.go`** — Remove `canUseSg`/`CanUseSg`, `sgUsable`/`sgUsableOnce`, `sync`/`runtime` imports; simplify `execIncusCommand`/`execIncusCommandContext` to always use `sh -c`; simplify `buildIncusCommand` to return a string
- **`internal/container/manager.go`** — Simplify `Available()` to always run `incus` directly; remove `runtime` import
- **`internal/cli/images.go`** — Remove `CanUseSg` conditional, always run `incus` directly
- **`internal/container/storage.go`** — Update comment
- **`tests/support/helpers.py`** — Remove `_has_sg()`, always use `incus` directly
- **`Makefile`** — Remove `sg` fallback in integration test targets
- **`tests/integration/test_sg_fallback.py`** — Removed (no longer relevant)
- **`CHANGELOG.md`** — Updated entry

## Test plan

- [ ] `go build ./...` compiles
- [ ] `golangci-lint run` passes (0 issues)
- [ ] `ruff check tests/` passes
- [ ] All existing integration tests pass (no behavior change — sg was already unused in CI)


mensfeld:
Addressed Copilot feedback in d74bbbf:

1. **IncusGroup unused** — Removed `IncusGroup` variable and the `group` parameter from `Configure()`. The `[incus].group` config field is kept in the config struct for backwards compatibility (won't break existing config files) but is no longer consumed.

2. **Use exec.Command("incus", ...) directly** — Valid suggestion but out of scope for this PR. The `sh -c` path is the same approach macOS always used and `shellQuote` handles escaping. Switching to direct exec would touch all command-building logic — better as a dedicated follow-up.

3. **Makefile preflight check** — Added `incus info` preflight check to both `integrations` and `integrations-debug` targets with a clear hint about re-logging after group changes.

Issue #354, opened 2026-05-04
Threat-event audit streaming (JSONL → stdout/file/syslog)


ChrisJr404:
Hi @mensfeld — long-time fan of the active-defense framing in this project.

One thing I noticed is that detection/containment events happen, but there's no first-class way to *export* them off-host. The agent detects something, takes action, and the evidence lives in stderr or wherever the operator is watching. For SIEM/SOAR plumbing — which is where I'd expect a lot of users to want to bolt this in — we'd want a structured stream.

Rough proposal:

- New `--audit` flag (and matching env var like `INCUS_AGENT_AUDIT=stdout|file:/var/log/...|syslog:user`).
- Emits one JSON object per line, schema something like:
  ```json
  {"ts":"2026-05-04T12:34:56.789Z","event":"threat.detected","agent_id":"abc","action_taken":"halted","detector":"network.outbound.exfil","details":{...}}
  ```
- `event` taxonomy starting with `threat.detected`, `threat.contained`, `agent.spawned`, `agent.terminated`, `policy.violation`. Open to extending.
- File sink uses an atomic-rename on rotation so log shippers like vector / fluent-bit don't truncate mid-write.
- Syslog sink uses RFC 5424 over UDP/TCP/TLS with structured-data block for the JSON body.

Happy to start narrow (just stdout-JSONL for the first PR, file/syslog as follow-ups) if that's a more reviewable shape. Two questions before I write code:

1. Do you want the audit stream to live in a separate process / daemon, or always in-process with the agent?
2. Any preference on structured log lib? `tracing-subscriber` with a custom layer feels natural for the in-process case, but if there's a project convention I'd rather follow it.

If the direction is good in general, I can put up a draft PR for the stdout-JSONL slice and we iterate from there.

mensfeld:
Makes sense. Please give me 24-48h to think about it and will get back to you 🙏

mensfeld:
Implemented in #362 (merged 2026-05-26). The  command provides structured JSONL streaming of threat events to stdout/file/syslog as described in the proposal.

PR #362, opened 2026-05-05
feat: threat-event audit streaming via in-container collector

ChrisJr404:
Refs #354.

Adds `coi audit` for emitting JSON-line audit records from a sandbox session.
Dump mode reads the host-side log the existing security monitor already
writes; `--follow` spawns a small in-container collector and streams live
events on stdout.

## Approach

The collector is a POSIX-sh agent (`internal/audit/agent/agent.sh`) embedded
into the binary via `go:embed`. On `--follow`, the host pushes the script
into the running container with the existing `incus file push` helper, then
runs it under `incus exec` and decodes each JSON line on stdout.

The agent picks the best telemetry source available:

1. **auditd** — `tail -F /var/log/audit/audit.log` if readable. Auditd `type=PATH` records become `type=file` events; everything else is forwarded as `type=audit` with the raw record preserved.
2. **syslog / auth.log fallback** — when auditd isn't installed, `tail -F` is run against `/var/log/syslog`, `/var/log/auth.log`, and `/var/log/messages`. Each line becomes a `type=audit` event.
3. **`ss -tunp` snapshots** every 5 s. Connections that weren't in the previous snapshot are emitted as `type=net` with proto/state/local/peer/pid/comm.
4. **`ps -eo pid,ppid,comm,args` snapshots** every 2 s. New pids since the previous tick become `type=exec` events.

No eBPF, no daemon install, no new Go deps. The agent runs only for the
lifetime of the `coi audit --follow` invocation.

## Event shape

One JSON object per line on stdout:

```json
{"ts":"2026-05-05T12:34:56.789Z","sessionId":"coi-abc-1","container":"coi-abc-1",
 "type":"exec","pid":42,"ppid":1,"comm":"curl","args":"curl https://example.com"}
{"ts":"2026-05-05T12:34:57.012Z","sessionId":"coi-abc-1","container":"coi-abc-1",
 "type":"net","proto":"tcp","state":"ESTAB",
 "local":"10.0.0.2:54321","peer":"169.254.169.254:80","pid":99,"comm":"python3"}
{"ts":"2026-05-05T12:34:58.110Z","sessionId":"coi-abc-1","container":"coi-abc-1",
 "type":"file","path":"/etc/shadow","op":"NORMAL"}
{"ts":"2026-05-05T12:34:58.111Z","sessionId":"coi-abc-1","container":"coi-abc-1",
 "type":"audit","msg":"collector.started","raw":"sources:auditd ss ps"}
```

The four `type` values are `exec`, `net`, `file`, `audit` — declared as
constants in `internal/audit/types.go` and validated by the parser.

## CLI

```sh
# Auto-detect container, dump host-side audit log
coi audit

# Specific container, dump host-side audit log
coi audit coi-abc-1

# Live in-container collector, JSONL on stdout
coi audit coi-abc-1 --follow

# Read a saved JSONL file (re-stamp sessionId / pretty-print)
coi audit --file ./session.jsonl | jq -c 'select(.type=="net")'
```

Container resolution reuses the same 3-tier logic as `coi monitor`
(positional arg / `COI_CONTAINER` / current-workspace lookup).

## Privacy + perf

Every event stays on the host running `coi audit` — nothing leaves the
machine unless the user pipes it elsewhere.

Measured idle overhead on Linux 6.19, sh+awk+sed+ss+ps:

- ~4.5 MB total RSS across the agent and its tail/awk helpers
- ~0.0% CPU when the container is quiet (workload above the noise floor
  was an extra ~0.3% during a `find /` test)

## Tests

- `internal/audit/parser_test.go` — 9 test functions (8 ParseLine cases + 3 StreamEvents cases): empty/whitespace, non-JSON stderr fallback, exec/net/file/audit decoding, missing/unknown type, malformed JSON, decoration + parse-error filtering, emit-error abort, 200KB lines.
- `internal/audit/proctree_test.go` — 5 test functions: ps snapshot parsing, new-pid diff, no-change diff, pid-reuse limitation case, empty-input handling.
- `internal/audit/agent_integration_test.go` — Linux-gated integration tests that run the real `agent.sh` against a fixture audit.log: asserts auditd PATH lines turn into `type=file` events with the right path and that every emitted line is valid JSON.

`go build ./...`, `go vet ./...`, and `go test -race ./internal/audit/... ./internal/cli/...` all pass.

## Limitations / non-goals for v1

- **No eBPF.** A bpftrace shim was considered and dropped — too much
  surface area for a first cut. Adding one in a follow-up would only need
  to plug a new emitter into agent.sh.
- **File events are coarse.** They come from auditd `type=PATH` records,
  not inotify, so coverage depends on the host's audit rules. When auditd
  isn't present at all, file events are absent (the syslog fallback
  doesn't synthesise them).
- **Pid-reuse blind spot.** The ps-diff emitter keys on pid; if pid N
  recycles within one tick, the new process is hidden. Documented in
  `proctree_test.go` (`TestDiffNewPIDs_PIDReuse`).
- **Polling not fsnotify.** The host-side `--follow` against the existing
  audit JSONL polls every second instead of pulling in fsnotify. Audit
  records are low-frequency (one per monitor poll, default 5 s) so the
  delta is invisible.
- **No audit-rule install.** If the operator wants `/etc/shadow` reads
  reported they still need to set up `auditctl` rules inside the
  container themselves; the agent only reads what auditd already logs.


mensfeld:
@ChrisJr404 great work on this!

One thing (if I'm reading it right): when the agent crashes or is killed inside the container, it looks like nothing happens loudly on the host side - the stream just goes quiet. Any chance we could detect that and restart, or at least surface it somehow?

The case I'm thinking about: a malicious actor inside the sandbox notices the agent is watching and kills it. It'd be great to at least notice when our visibility goes dark, vs. assuming the workload just got quiet.

ChrisJr404:
Good call, the silent-failure case is real. Off the top of my head:

- agent-side: emit a periodic heartbeat event (separate stream type, every 10s or so)
- host-side: track last-seen time per agent connection. If we miss N heartbeats in a row, log a warning and either reconnect (if the container is still up) or surface it on the overview TUI.

That keeps this PR scoped to the audit streaming itself and lets the heartbeat land as a follow-up since it touches both the agent protocol and the host-side state. Want me to file that as a separate issue/PR after this one merges, or fold it in here?

mensfeld:
If you are ok folding here it would be great, otherwise i have to create a separate card, track it etc. :pray:

ChrisJr404:
Folded the heartbeat in — pushed `1bf4c09` + `84b49c9`.

**Wire shape:** agent emits `{"type":"heartbeat","seq":N,"sources":"auditd ss ps"}` every 10 s. Sequence number is monotonic per session so dropped heartbeats are visible. Cadence is overridable via `COI_AUDIT_HEARTBEAT_INTERVAL`.

**Host side:** new `audit.HeartbeatWatcher` keeps the most recent heartbeat per session and flips a session to stale at 35 s (3 missed + grace), checked on a 5 s ticker. When a session goes stale or recovers, the watcher fires a callback. The `coi audit --follow` runner wires it to surface stale events two ways:

- stderr for the operator: `[audit] WARNING agent silent on coi-abc-1 for 36s (last heartbeat 2026-05-05T12:34:46Z)`
- inline as a `type=audit msg=agent.stale` JSONL event so any SIEM / future overview consumer that already reads the stream picks it up the same way it picks up everything else. `agent.alive` clears the warning when heartbeats resume.

Watcher does not auto-kill the connection — operator decides. Stdout writes from the collector and the watcher are serialised behind a shared mutex so the JSONL stream never interleaves.

**Tests:** unit coverage on the watcher (first-seen, stale-after-threshold, no-duplicate-stale, recovery, multi-session independence, empty-id no-op, forget) plus an integration test that runs the actual `agent.sh` with a 1 s heartbeat interval and asserts seq=0 then seq=1 land on stdout. `go test -race -short ./internal/audit/... ./internal/cli/...` is green; the only failure in `./...` is the pre-existing `TestCheckKernelVersionIntegration` on Kali kernels (`6.19.11+kali-amd64` — unrelated).

**Fold-in fix:** the merge of #360 into this branch left two references to `container.CanUseSg` / `container.IncusGroup` in `internal/cli/audit.go` that no longer compile. Replaced with the direct `sh -c incus ...` path the rest of the package uses post-#360 — that's why the diff includes `newIncusPipedCommand`.

README section on the audit stream now documents the heartbeat behavior, the env knob, the defaults, and the shape of stale/alive events.

mensfeld:
Amazing. I will do my best to re-review within upcoming days. THANK YOU! :pray:  :heart:

mensfeld:
Reviewed it once more and since im not an expert in OS ev ents i wanted to check with claude if some of those could be taken from the host instead of the container (to prevent blindness) and here is what I've got:

| Source | Host-side approach | Extra install on host? |
|---|---|---|
| auditd records | Read the container's `/var/log/audit/audit.log` via its rootfs path on the host, with inotify | None — `inotify` is a syscall, pure Go |
| syslog / auth.log | Same — read via container rootfs path with inotify | None |
| Network connections | Read `/proc/<container-init-pid>/net/tcp`, `tcp6`, `udp`, `udp6` from the host. Those files are namespaced, so reading them via the init pid gives you the container's view | None — just file reads |
| Process list (poll) | Walk `/proc` on the host, filter pids whose `/proc/<pid>/ns/pid` matches the container init's | None — just file reads |
| Process events (push) | Open a netlink socket with `NETLINK_CONNECTOR` + `PROC_EVENTS`, filter by pid namespace | None — pure Go via `syscall`/`x/sys/unix`. Needs `CONFIG_PROC_EVENTS=y` in the kernel (universal on stock distro kernels) |
| File events (push, without auditd) | `fanotify` with `FAN_MARK_FILESYSTEM` on the container's rootfs path | None — pure Go syscall. Needs `CONFIG_FANOTIFY=y` (universal) |
| Namespace entry (if needed) | `setns(2)` syscall directly from Go | None — pure Go via `x/sys/unix.Setns` |

That said I am ok shipping this as long as we (I can help!) open (no rush) an issue to move those that are movable to the host. That way the tamper evidence becomes much more solid.
 
WDYT?

mensfeld:
@ChrisJr404 hope you don't have the feeling im neglecting this PR. I just have years of malicious research and I've seen stuff :roll_eyes: so i want to make sure that we make it as bulletproof as possible.

ChrisJr404:
No worries on the timing — appreciate you stress-testing this rather than just nodding it through.

Quick reactions on the host-side options:

* auditd / syslog read from the host via inotify on the container rootfs is the cleanest of the bunch. No extra deps and the container can't tamper with the host's view of those files.
* `/proc/<init>/net/*` via the host's view is a nice trick, hadn't clocked that those are namespaced through the init pid.
* `proc-events` netlink is pure Go and pretty universal on stock kernels.
* `fanotify` with `FAN_MARK_FILESYSTEM` is also pure Go, good for the file events tier.
* `setns` is the right move when we actually have to enter the namespace.

I'm OK with shipping this PR as-is and opening a follow-up tracking issue for the move-to-host migration. The heartbeat watcher already gives operators a "we went blind" signal, so the in-container observer staying for now is an explicit tradeoff rather than a hidden risk, and the host-side sources slot in cleanly under the existing `sources` enum without breaking the wire format.

Want me to file the follow-up issue now or after this lands?

mensfeld:
> Want me to file the follow-up issue now or after this lands?

Feel free to create an issue now, but ofc this will be merged and released, and we can gradually migrate sub-parts with the same reporting API (you good with that?)

mensfeld:
All good here, I will merge it soon. Had some family issues sorry for the delay :/

mensfeld (approved):

mensfeld:
Ok, fixing linting issues.

mensfeld:
merged :) will go live with the upcoming release!

PR #432, opened 2026-06-04
fix(monitor): walk host /proc to collect container process list

mensfeld:
## Summary

- Replace \`incus exec ps aux\` with a host-side \`/proc\` walk filtered by PID namespace
- An attacker inside the container can no longer hide processes by manipulating their \`/proc\`, replacing \`ps\`, or pausing \`incus exec\`
- PPIDs are now accurate — \`ps aux\` always returned PPID=0
- **No in-container fallback**: if the host walk fails, \`Available=false\` is returned rather than silently falling back to the tamper-able \`incus exec\` path

## How it works

1. Resolve the container's init PID via \`GetContainerInitPID\` (from PR #430)
2. Read the init's PID namespace inode from \`/proc/<initPID>/ns/pid\` (looks like \`pid:[4026532456]\`)
3. Walk host \`/proc\` — for each numeric directory, compare its \`/proc/<pid>/ns/pid\` to the container's namespace inode; keep only matches
4. For each matched PID, read \`/proc/<pid>/status\` (Name, PPid, Uid) and \`/proc/<pid>/cmdline\` (null-delimited args)

## New testable helpers

- \`parseStatusFields(status string) (name string, ppid, uid int)\` — parses \`/proc/<pid>/status\` content
- \`parseCmdline(data []byte) string\` — converts null-delimited cmdline bytes to a space-joined string

Both covered by unit tests in \`process_test.go\`.

## Integration tests

New \`TestProcessHostProcWalk\` class in \`test_security_monitoring.py\`:
- \`test_env_scan_detected_via_host_proc_walk\`: injects \`exec -a 'env' sleep 30\`; verifies WARNING threat detected via host proc walk
- \`test_reverse_shell_detected_via_host_proc_walk\`: injects \`exec -a 'python -c socket.socket' sleep 30\`; verifies CRITICAL + container kill

## Test plan

- [ ] CI green
- [ ] Copilot review addressed

### DIFF
--- a/internal/monitor/daemon.go
+++ b/internal/monitor/daemon.go
@@ -77,6 +77,13 @@ func (d *Daemon) run() {
 				continue
 			}
 
+			// Surface any per-source collection errors so they appear in logs.
+			for _, collErr := range snapshot.Errors {
+				if d.config.OnError != nil {
+					d.config.OnError(fmt.Errorf("collection: %s", collErr))
+				}
+			}
+
 			// Detect threats
 			threats := d.detector.Analyze(snapshot)
 			snapshot.Threats = threats
--- a/internal/monitor/process.go
+++ b/internal/monitor/process.go
@@ -3,15 +3,139 @@ package monitor
 import (
 	"context"
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
 
 	"github.com/mensfeld/code-on-incus/internal/container"
 )
 
-// CollectProcessStats collects running processes from the container
+// CollectProcessStats collects running processes from the container.
+// It first attempts to walk the host /proc filtered by the container's PID
+// namespace, which is tamper-resistant because it runs outside the container.
+// If that fails it falls back to incus exec ps aux.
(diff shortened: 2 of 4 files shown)


mensfeld:
The 4 Copilot inline comments (from the initial commit) all refer to code that was subsequently replaced:

1. **Empty process guard**: Added in a later commit — `collectProcessesViaHostProc` now returns an error when `len(processes) == 0`.
2. **PID-namespace inode comparison**: Replaced entirely with cgroup-membership matching (`processBelongsToContainerCgroup`), which is world-readable and covers nested namespaces without ptrace.
3. **`incus exec` fallback**: Removed — the function returns `Available: false` on failure instead.
4. **Numeric UID in User field**: An inline comment is present on that line explaining it is the host-side UID.

The current code also strips trailing `.scope`/`.service` cgroup path components (fixed in the most recent commit) so processes started via `incus exec` — which land in a different sub-scope than init — are correctly matched.

Issue #530, opened 2026-07-02
Colima/Lima auto-detect disables UID shift but never sets raw.idmap, so workspace writes fail when host UID != 1000


technicalpickles:
## What's going on

COI detects Colima/Lima environments (virtiofs mounts in `/proc/mounts`) and disables UID shifting for them, which is the right call — the VM already handles UID mapping. But the code stops there. It never sets `raw.idmap`, so once shift is off, the container's `code` user (UID 1000) has no way to write into a workspace owned by whatever UID your host user actually is. For most Mac users that's 501, not 1000, so this isn't an edge case — it's the default case.

## How I found it

I was testing coi on an OrbStack Ubuntu 24.04 VM as a Colima/Lima stand-in (Incus 7.2 via Zabbly, `coi build` ran clean). First `coi run` against a workspace:

```
$ coi run -- bash -c 'touch /workspace/test.txt'
...
Command exited with code 1
```

`coi health` correctly flags the environment (`Operating system : Ubuntu 24.04.4 LTS (arm64) [colima]`), and inside the container `id` shows `uid=1000(code)`. The workspace directory is owned by uid 501, mode 755 — readable, not writable by anyone else. Colima and Lima both create their default guest user at the same UID as the host Mac account, so this isn't OrbStack-specific. It'll hit real Colima/Lima setups the same way, any time your Mac's primary user isn't UID 1000.

## Root cause

`internal/session/setup.go`, roughly lines 276-303:

```go
disableShift := opts.DisableShift
if !disableShift && isColimaOrLimaEnvironment() {
    disableShift = true
    opts.Logger("Auto-detected Colima/Lima environment - disabling UID shifting")
}

useShift := !disableShift
hostUID := os.Getuid()

if !disableShift && hostUID != container.CodeUID {
    // raw.idmap gets set here — but only when disableShift is false
    ...
} else if disableShift {
    // Colima/Lima path: shift is off, raw.idmap never gets set
    ...
}
```

The `raw.idmap` branch only runs when `!disableShift`, so it's unreachable the moment Colima/Lima auto-detection kicks in. Worth noting this was a deliberate scope cut, not an oversight: #166 and #221 covered the same underlying bug for the general/CI case, fixed in #226 (commit 7cf8ec8), and that PR's own plan says `// Colima/Lima handling (unchanged) ...`. So the Colima/Lima path just never got the fix.

## Suggested fix

[...]

PR #552, opened 2026-07-05
fix(run): attach disk devices pre-start so the isolation fallback covers virtiofs workspaces (#534)

### LINKED ISSUE #534: Virtiofs-backed workspaces fail to hot-mount under security.idmap.isolated, no fallback like the container-start path has
## What's going on

`coi run` starts the container first, then hot-mounts the workspace into it as a separate step. If the workspace lives on a virtiofs mount, that hot-mount fails outright and takes the whole run down with it:

```
Error: Failed to start device "workspace": Required idmapping abilities not available
Cleaning up container coi-a96cacbd-1...
Error: configure-container: failed to mount workspace: exit status 1
```

Root cause: coi sets `security.idmap.isolated=true` on every container (real hardening, gives each container its own isolated UID namespace), and that requires the underlying filesystem to support kernel idmapped mounts. The VM's own filesystem (btrfs, tested) supports it fine. Virtiofs, at least as OrbStack implements it, doesn't.

coi already has a fallback for isolation failures, but it only covers the container itself failing to *start* under isolation (`StartWithIsolationFallback` in `internal/container/commands.go`, unset `security.idmap.isolated` and retry). That fallback runs during the `launch-container` phase, before the workspace is mounted at all. The workspace device gets added afterward, in `configureContainerRunPhase` / `applyWorkspaceMounts` (`internal/cli/run.go:356`, `mgr.MountDisk("workspace", ...)`), as a live hot-add to a container that's already running and already isolated. There's no retry path for a device failing to hot-mount, so the very first idmap-incompatible device just fails hard and the run tears itself down.

## How I found it

Reproduced end to end on the OrbStack Ubuntu 24.04 VM (`coi-test`) I've been using as a Colima/Lima stand-in. `/Users` on that VM is virtiofs (the Mac share); `/home` is native btrfs.

```
$ cd /Users/technicalpickles/coi-virtiofs-workspace-test
$ coi run -- bash -c 'echo hello from container'
...
Waiting for container to be ready...
Remapping user code from UID 1000 to 501...
Mounting workspace /Users/technicalpickles/coi-virtiofs-workspace-test -> /workspace...
Error: Failed to start device "workspace": Required idmapping abilities not available
Cleaning up container coi-a96cacbd-1...
Error: configure-container: failed to mount workspace: exit status 1
```

Same project directory on the VM's native filesystem (`/home/...`) works fine. So this isn't about the workspace content, it's specifically about the mount's filesystem.

## Practical implication

Since the docs say the model is "shell into your Linux VM and run coi from inside it," it's natural to reach for whatever's convenient, including a project checked out on the Mac and visible through the VM's shared-folder mount. That path is dead on arrival right now: workspaces need to live on the VM's own filesystem, not just be reachable through the Mac share, and nothing in the tool says so up front. You just get an idmap error mid-launch.

## Possible directions

- Extend the existing isolation fallback to cover device hot-mount failures too, not just container-start failures: if mounting the workspace fails with an idmap error, unset `security.idmap.isolated` on the running container and retry the mount, same pattern `StartWithIsolationFallback` already uses for start.
- At minimum, detect the workspace filesystem type before attempting the mount and fail with a clear message ("workspace is on a virtiofs mount, which doesn't support idmapped mounts under security.idmap.isolated; move it to the VM's native filesystem or disable isolation") instead of the raw Incus device error.

## Environment
- COI v0.9.0 (`v0.9.0-88-g2f15f5e`)
- Incus 7.2 (Zabbly repo)
- Host: OrbStack Ubuntu 24.04 VM (macOS), standing in for Colima/Lima

mensfeld:
Closes #534. Built on the lima-smoke lane (#550), which is exactly the environment that reproduces this.

## Root cause (confirming the reporter's diagnosis)

`security.idmap.isolated=true` (set on every COI container) requires each disk device's filesystem to support kernel idmapped mounts; virtiofs (as OrbStack/Colima implement it) doesn't. The existing fallback (`StartWithIsolationFallback` — unset + retry) only covered the container **start**. `coi run` hot-mounted the workspace **after** start → no fallback → hard failure:

```
Error: Failed to start device "workspace": Required idmapping abilities not available
```

**Key finding:** the `coi shell` path never had this bug — it attaches all devices *before* start (setup.go: "Add disk devices BEFORE starting container"), so the failure surfaces at start where the fallback already handles it. `coi run` was the outlier.

## The fix — make run mirror shell (right altitude, no new fallback machinery)

The pre-start hook built for #530's `raw.idmap` now also attaches **all** disk devices (workspace, additional mounts, protected-path + secret-mask overlays) to the stopped container. Mount/socket parsing + trust gating move before launch (pure host-side computation). Consequences:

- An idmap-incompatible device fs now fails the **start**, where the existing fallbacks take over: non-ephemeral → unset isolation + retry; **ephemeral → recreate without isolation, and the pre-start hook re-runs on the fresh container, re-applying the UID mapping and every device** (that re-run semantics already existed for #530).
- Isolation stays ON everywhere it works — it is only dropped on hosts that can't support it, same contract as the existing start fallback.
- Reused persistent containers restart via `StartWithIsolationFallback` too (creation-time devices materialize at start — same failure class).
- A failed fresh launch best-effort-deletes the half-created container (stopped ephemeral containers are only auto-deleted after having run).

Also addresses the reporter's alternative suggestion implicitly: the reactive "unset on the running container and retry the hot-mount" wouldn't work — idmap changes only apply at (re)start, and restarting an ephemeral container deletes it. Pre-start attachment is the only ordering where the failure lands on a recoverable path.

## Regression test (in the lane built for it)

`tests/run/test_virtiofs_workspace.py` — creates its workspace **on the virtiofs mount** and asserts `coi run` succeeds end-to-end (write visible on the mount, in-container writer uid 1000). Runs in the **lima-smoke** lane via `COI_TEST_VIRTIOFS_DIR` (the guest's mounted host `$HOME`); self-skips anywhere without virtiofs. `tests/run` is already a matrix path, so the coverage guard is satisfied with no matrix changes.

## Gates

build, vet, gofmt, cli+container+session unit suites, golangci-lint v2.12.2 → 0 issues, ruff, yaml parse, check-ci-coverage — all clean. The decisive verification is the lima-smoke lane on this very PR: it now runs the #534 repro with the fix in place.


mensfeld:
CI green (26/26) — verification evidence from the lima-smoke lane on this PR:

```
tests/run/test_virtiofs_workspace.py::test_run_works_with_virtiofs_workspace PASSED (26.8s)
============================== 7 passed in 48.54s ==============================
```

The #534 repro ran for real: workspace created **on the virtiofs mount** inside the Lima guest, `coi run` succeeded end-to-end, write visible on the mount, in-container writer uid 1000. The 26.8s runtime (vs ~3s for the native-fs sibling tests) is consistent with the isolation fallback engaging (start fails under `security.idmap.isolated` → ephemeral recreate without isolation → start).

Iteration log for review context:
1. `/dev/kvm` udev race in the lane's KVM check (infra, de-raced with settle+retry+chmod fallback).
2. lima lane green, but `misc-config` caught a **real regression** in the refactor: moving mounts pre-start moved `chattr +i` earlier while its release stayed in the configure-phase teardown, which never registers on early failures (e.g. invalid limits config) → leaked immutable flags → pytest tmpdir cleanup EPERM. Fixed by pairing release with acquisition: the launch phase's teardown removes the flags and is returned even alongside a launch error.
3. All green.

Issue #551, opened 2026-07-05
lima-smoke: cover the uid-1000 + Colima/Lima branch (detection-gated shift-off path)


mensfeld:
Follow-up to the lima-smoke CI lane (#543 / #550).

## Gap

`decideUIDMapping` has one branch no CI lane covers: **host uid == 1000 (matches container code uid) while Colima/Lima is detected** → `useShift=false`, no `raw.idmap` — the historic Colima happy path (the 0.6.0-era "disable shift under Lima" behavior). It is also the *only* branch where detection changes observable behavior: at any uid mismatch, the raw.idmap decision is identical with or without detection, so native lanes (uid 1001) and the current lima-smoke tests (guest uid 1001) both exercise the mismatch path only.

## Idea

The Ubuntu cloud image in the Lima guest ships an `ubuntu` user with **uid 1000**. Add a lima-smoke step that runs one `coi run` write-test as that user:

- `sudo usermod -aG incus-admin ubuntu` (or rely on the socket chmod the lane already does)
- workspace on the guest-native fs owned by `ubuntu:1000`
- `sudo -iu ubuntu env COI_BINARY=/usr/local/bin/coi ... coi run --workspace ... -- bash -c 'echo ok > /workspace/f'`
- assert the write lands and the in-container writer is uid 1000, with **no raw.idmap set** and **shift off** (the detection log line "Auto-detected Colima/Lima environment - disabling UID shifting" should appear).

## Caveat / why this is not in v1

It is not obvious this branch even *works* today: with shift off and no idmap, the unprivileged container's uid 1000 maps to a high host uid, so writes to a 1000-owned workspace may fail — i.e., this probe may surface a real product bug in the uid-match-under-Lima path rather than pass. That's the point of adding it, but it needs investigation first, not a blind blocking assert.

_From the #550 plan's deliberate v1 scoping._

mensfeld:
Closing — the actionable part is already covered, and the requested end-to-end probe can't be done meaningfully in CI.

**The decision is unit-tested.** `decideUIDMapping(1000, 1000, disableShift=false, colima=true) → shift=false, idmap=""` is asserted in `internal/session/uid_mapping_test.go` ("match, colima → no shift, no idmap"), so the uid-1000 + Colima/Lima branch is regression-safe at the logic level.

[...]

PR #554, opened 2026-07-05
Fix OrbStack virtiofs mount misdetected as Colima/Lima

### LINKED ISSUE #553: OrbStack's virtiofs mount is misdetected as Colima/Lima, disabling UID shift and breaking /workspace writes
### What happened

Testing a custom profile on OrbStack (not Colima/Lima), writes to `/workspace` from inside a container failed with `EACCES`. A follow-up `sudo chown -R $(id -u):$(id -g) /workspace` run from inside the container also failed, with `Operation not permitted` on most files and `Read-only file system` on the protected paths (`.husky`, `.vscode` etc, which is expected/intentional and unrelated to this bug).

### Root cause

`isColimaOrLimaEnvironment()` (`internal/session/setup_helpers.go:16`) detects Colima/Lima by checking for `virtiofs` in `/proc/mounts`. OrbStack also shares the host filesystem into its VM via virtiofs (`mac on /Users type virtiofs`, etc), so this check false-positives on OrbStack too.

When it fires, `setup.go` disables UID shifting entirely (`useShift = false`) and skips the `raw.idmap` fallback too, since that branch is gated on `!disableShift`. There's also no way around it from config: `disable_shift` (`internal/config/config.go:193`) is a plain bool that can only force shift *off*. It can't force shift back *on* when auto-detection wrongly disables it, because a config value of `false` is indistinguishable from "not set."

### Confirmed with

- The container's `workspace` disk device has no `shift` and no `raw.idmap` set (checked via `incus config show`/`incus config device show`).
- Every file under `/workspace` shows up owned by `65534:65534` (`nobody:nogroup`) inside the container, because the real host UID (which I'd set to match `code_uid` in `~/.coi/config.toml`, needed for OrbStack) falls outside the container's isolated idmap range and can't be resolved.
- Manually stopping the container and setting `shift=true` on the workspace device fixes it completely: ownership flips to the correct `code:code`, and writes work fine afterward.

```
incus stop <container>
incus config device set <container> workspace shift=true
incus start <container>
```

Not a durable fix, coi recreates the device without shift on the next `coi shell`/`coi run`, but it confirms OrbStack's virtiofs mounts do support idmapped shift mounts just fine, this is purely a detection problem, not a fundamental OrbStack limitation like it is for actual Colima/Lima.

### Suggested fix

Tighten the Colima/Lima detection so it doesn't fire on OrbStack. A couple of directions, happy to send a PR for whichever you'd prefer:

- Check `uname -r` for `orbstack` (OrbStack's guest kernel literally reports as `7.0.11-orbstack-...`), or check for `/opt/orbstack-guest`, and exclude those before falling back to the virtiofs check.
- Or add a config knob that lets a user force shift on regardless of auto-detection (e.g. a tri-state `disable_shift`, or a separate `force_shift` field), so anyone on a lesser-known virtiofs-based VM (or whatever this heuristic gets wrong next) isn't stuck waiting on a detection fix.

Also noticed the same `isColimaOrLimaEnvironment()` false-positive gates the Bedrock-in-Colima validation check at `setup.go:113`. Lower stakes (it's just an extra validation path), but same root cause.

technicalpickles:
## Summary

- `isColimaOrLimaEnvironment()` treats any `virtiofs` mount as proof of Colima/Lima and disables UID shifting (and the `raw.idmap` fallback) accordingly. OrbStack also shares the host filesystem via virtiofs, so it trips the same check.
- With shifting disabled, `/workspace` ends up mounted with no UID translation at all: files show up owned by the overflow UID (`nobody:nogroup`) inside the container, and writes fail with `EACCES`.
- Fix excludes OrbStack up front by checking its guest kernel's release string (e.g. `7.0.11-orbstack-...`) before falling back to the virtiofs/lima-user checks.

Fixes #553

## Test plan

- Added `TestDetectColimaOrLima`, covering the genuine Lima virtiofs case, the `lima`-user case, the new OrbStack exclusion, and the negative case.
- Confirmed live on an OrbStack VM: a fresh `coi run --profile ollama-cloud` now mounts `/workspace` as the `code` user and writes succeed, with no manual `incus config device set ... shift=true` workaround needed.


mensfeld (commented):
Reviewed in depth — **this is correct, well-scoped, and composes cleanly with the recent #530/#534 work**. Verified locally on a checkout of this branch: the new table test passes, the full `internal/session` suite passes, gofmt/vet/golangci-lint (v2.12.2) all clean, and the branch is already based on current master.

## Why the fix is right

- The root-cause analysis holds on current master. Post-#540, a **uid mismatch** would still have been rescued by `raw.idmap` even under misdetection — but your setup has `code_uid` matching the host uid, which lands in the **uid-match + detection-fires** branch of `decideUIDMapping`: shift forced off, no idmap, no mechanism at all → `nobody:nogroup`. Exactly what you observed.
- Excluding OrbStack from detection is the correct semantic fix, not a workaround: your `shift=true` experiment proves OrbStack's virtiofs supports idmapped mounts, so OrbStack simply *isn't* in the class this heuristic exists for (Colima/Lima, where shift genuinely can't work). After exclusion, OrbStack gets the normal path: shift on uid-match, `raw.idmap` on mismatch — both mechanisms OrbStack supports.
- Blast radius checked — all three call sites behave correctly under the exclusion: the UID-mapping decision (main fix), the shift-disabled log line (cosmetic), and the Bedrock-under-Colima validation gate at `setup.go:114` (your issue's side note — fixed transitively).
- The `/proc/mounts` read-error path subtly improves: previously an unreadable mounts file returned `false` outright; now the `USER=lima` fallback still gets a say. Strictly better.
- Splitting the pure `detectColimaOrLima` for unit testing matches the pattern we use for `decideUIDMapping`. 👍

## Notes (non-blocking)

1. **The residual class is tracked**: your bug is one instance of the "uid-match while detection fires → nothing engages" branch (#551 tracks it for *genuine* Lima, where it's likely still broken — there's no shift to fall back on there). This PR correctly removes OrbStack from that broken branch rather than trying to fix the branch itself.
2. **CI's lima-smoke lane is the perfect counter-test here**: it runs COI inside a genuine Lima VM (virtiofs mount, non-OrbStack kernel), so it verifies your exclusion doesn't loosen real Lima detection. Watching it go green on this PR is the end-to-end proof.
3. **CHANGELOG**: this repo gives every user-facing fix an entry in the 0.10.0 `### Fixed` section — happy to add one for you at merge time if you'd rather not push another commit (suggested angle: "OrbStack is no longer misdetected as Colima/Lima (#553) — its guest kernel is excluded before the virtiofs check, so UID shifting stays available; OrbStack's virtiofs supports idmapped mounts, unlike Colima/Lima's").

Thanks for the third excellent find-diagnose-fix in this area — the detection heuristic is meaningfully better tested now than before.

PR #555, opened 2026-07-05
Extract shared host-VM detection into internal/vmhost

technicalpickles:
Stacked on #554 — this PR's diff includes that one's commit until it merges. The two commits specific to this PR are the last two in the log (`refactor(session): name the shift-detection helper after what it checks` and `refactor(vmhost): extract shared host-VM detection out of session and health`).

## Summary

- `session`'s shift-detection helper was named after two specific VM providers (`isColimaOrLimaEnvironment`), which stopped fitting once #554 added an OrbStack exclusion to it. Renamed to `hostHandlesUIDMapping`/`detectHostUIDMapping` to name what it actually decides.
- Turned up a second, independent copy of the same heuristic in `internal/health/checks.go` (`isColimaEnvironment`), used to label `coi health`'s environment field and tailor nft troubleshooting messages. It had drifted from `session`'s copy and would still misreport an OrbStack host as "colima."
- Extracted both into a new `internal/vmhost` package with a `Kind` enum (`KindUnknown`/`KindLimaLike`/`KindOrbStack`) rather than a bare bool, since `session` and `health` each read a different property of the same detection (`HandlesUIDMapping()` vs. a display label).

## Test plan

- `go test ./internal/vmhost/...`, `./internal/session/...`, `./internal/health/...` all pass in the Linux VM used to develop #554.


technicalpickles:
@copilot resolve merge conflicts

mensfeld:
@technicalpickles wont work that way for OSS but I will take care :pray:

mensfeld:
Rebased onto current master to resolve the conflicts (maintainer edit).

**What changed:** the branch was stacked on #554, which merged separately — so I rebased **--onto master**, dropping the now-redundant #554 commit (the session-side OrbStack fix is already in master) and keeping just the two refactor commits. Authorship preserved.

**Conflict resolution:** the two refactor commits predated the #530 `ConfigureUIDMapping` centralization, so `setup.go`/`setup_helpers.go` conflicted. Resolved by keeping master's `ConfigureUIDMapping`/`decideUIDMapping` and wiring the detection through the new `vmhost` package (`vmhost.Detect().HandlesUIDMapping()`, called once) — preserving both #530 (raw.idmap on UID mismatch) and #554 (OrbStack exclusion) behavior.

**Still worth merging (confirmed):** #554 only fixed the *session* copy. Master's `internal/health/checks.go` still has the old `isColimaEnvironment()` (bare virtiofs check, no OrbStack exclusion) at lines 54/409 — so `coi health` still misreports an OrbStack host as 'colima'. This PR fixes that (`vmhost.Detect() == vmhost.KindLimaLike`) **and** consolidates the two drifted copies into one package, preventing exactly the drift that made #554 miss the health copy.

Local: `go build/vet`, gofmt, golangci-lint v2 (0 issues), and the full `./internal/...` unit suite all pass, incl. the vmhost OrbStack test.

Issue #556, opened 2026-07-06
Git identity resolution only reaches coi's three built-in tools; custom tools are on their own


technicalpickles:
### What happened

Ran a session through the `coi-ollama-cloud` profile (uses `oh-my-pi`, a custom tool, not one of coi's three built-in tools). It committed as `foobar-bot <bot@foobar.local>`, a fully hallucinated identity, not close to my real name/email and but was derived from something in the repo

### Root cause

coi's git identity handling has two parts:

1. A guard (`SetupGitIdentityGuard`, `internal/session/setup_helpers.go:87`) that sets `user.useConfigOnly=true` globally in the container, so git refuses to guess identity from the OS user.
2. Instructions written into `~/SANDBOX_CONTEXT.md` telling the driving AI tool how to discover the real identity (SSH agent greeting, then `gh api user`, then `git log` on existing history, then ask the user), and explicitly telling it not to fabricate one.

Part 2 only reaches Claude Code, opencode, and coi's own `pi` tool, since it's injected via each tool's native auto-context mechanism (`ToolWithAutoContextFile`, e.g. `~/.claude/CLAUDE.md`). Any other tool, like `oh-my-pi` here, never sees the file or the instructions, even though `SANDBOX_CONTEXT.md` gets written to disk regardless.

So for a custom tool, the guard fires (good, no silent misattribution as the container's `code` user), but nothing tells the model what to do about it. In this case the model just made something up instead of asking, which is exactly the failure mode the instructions were written to prevent. It just never got to read them.

This is adjacent to #337 (which added the discovery instructions in the first place, scoped explicitly as "hints" for the driving tool) and #549 (generic host-credential seeding for tools coi doesn't know about, same underlying gap, different resource).

### Proposed direction

Move git identity resolution out of "instructions we hope the model follows" and into something coi actually does, independent of which tool is running:

- coi resolves identity itself during session setup (SSH agent lookup, then `gh api user`, then `git log` fallback) and runs `git config user.name`/`user.email` for real, the same way it already runs `SetupGitIdentityGuard`.
- A `[git]` config block with a default resolution chain, overridable with an explicit `name`/`email`, or a custom resolution command, the same shape as the `[[credentials]]` idea in #549, applied to identity instead of files.
- If resolution comes up empty, leave the guard in place so git fails loudly on commit rather than falling back to fabrication, same safety property as today, just without depending on a specific tool actually reading and following the instructions.

### Tradeoff

[...]

PR #562, opened 2026-07-06
fix(session): seed git identity from host config

fengjikui:
Refs #556

## Summary
- resolve a complete host-global git identity (`user.name` + `user.email`) before `coi shell` setup
- configure that identity inside the container immediately after `user.useConfigOnly=true`
- keep the existing fail-closed guard when either field is missing, and avoid project-local git config so an untrusted checkout cannot choose the container author identity
- expose the same identity plumbing on `ConfigureOptions` for orchestrators that already use the git guard

## Validation
- `go test ./internal/session -run 'TestSetupGitIdentity|TestSetupGitIdentityGuardUsesEscapedHome' -count=1`
- `go test ./internal/session -short -run 'TestSetupGitIdentity|TestSetupGitIdentityGuardUsesEscapedHome|TestContextFile_ForwardedEnvVars|TestSetupGitIdentityGuard' -count=1`
- `go test ./internal/config ./internal/tool -count=1`
- `git diff --check`

## Local platform notes
- `go test ./internal/cli -run 'TestResolveHostGitIdentity' -count=1` cannot compile on this macOS host because `internal/cli` pulls in Linux-only monitor code (`unix.SizeofInotifyEvent`, fanotify constants). The test should be exercised by the Ubuntu CI lane.
- A broader `go test ./internal/session ./internal/cli` attempt hit the same `internal/cli` macOS compile boundary plus unrelated pre-existing macOS session test failures around immutable degradation and long Unix socket paths.


mensfeld:
Oooooo that is neat!

mensfeld:
Will make a small change and ready to go :)

mensfeld:
Hey, thank you for the contribution. I made a small change: I made this configurable because some of the users (including myself) use this without the host git config so i made this configurable. I also aligned the sandbox context to have notion of all of this.

PR #573, opened 2026-07-08
Add generic credential catalog for third-party tool providers

### LINKED ISSUE #549: Generic host-credential seeding for containers (currently hardcoded per-AI-tool, no hook for third-party integrations like ollama)
### What I'm trying to do

I put together a custom profile that installs `ollama` alongside Claude Code, so `ollama launch claude --model <cloud-model>` can run inside a coi container against Ollama Cloud. That needs the container to have the same signed-in identity as the host: an SSH keypair `ollama` generates on first sign-in and ties to your ollama.com account. Same basic need as Claude needing `~/.claude/.credentials.json`.

For Claude/opencode/pi, coi already solves exactly this: `ToolWithConfigDirFiles` (`ConfigDirName()`, `EssentialConfigFiles()`, etc.) on the `Tool` interface handles copying host credentials into a fresh container, and pulling tool state back out on session save. That machinery is only reachable by the three tools registered in `internal/tool/registry.go` though. There's no way for a profile, or a third-party integration like ollama, to say "also seed this host path into every new container" without writing Go and rebuilding `coi` itself.

Right now I'm doing this by hand: `incus file push` the identity file into a running container after `coi shell`/`coi run` launches it, then re-`chown` it to whatever the container's local user's UID happens to be (can't just bind-mount it live, since the same logical user's UID isn't guaranteed to match between host and container). Works fine for one-off experimentation, but it's not the normal "just `coi shell` and it works" flow the rest of coi has.

### What might help

Something like a generic `[[credentials]]` section in profile config, parallel to the existing `[[mounts]]`:

```toml
[[credentials]]
host = "~/.ollama/id_ed25519"
container = "~/.ollama/id_ed25519"
mode = "0600"
```

that copies matching files into a freshly-created container the same way `setupCLIConfig` already does for the registered AI tools, just without requiring the source/target to be tied to a `Tool` implementation. Has to be a copy-then-chown-to-the-container-side-owner, not a bind mount, for the UID-mismatch reason above.

This would help anyone bringing a host credential or config file in for a tool that isn't (and doesn't need to be) one of coi's officially supported AI tools.

Not urgent on my end, just flagging it since I hit it directly. Happy to take a swing at implementing it if this direction is welcome, though I'd want your take on the config shape first.

technicalpickles:
## Summary

- New profile-level `[[credentials]]` config: reference a named catalog bundle (`bundle = "ollama"`) or declare an ad-hoc host/container file pair for anything not yet cataloged.
- `claude`/`opencode`/`pi`'s own credential config now comes from that same embedded catalog instead of being hardcoded per-tool in Go, so third-party providers like Ollama get the same treatment as the built-in tools.
- Ad-hoc entries get the same untrusted-project-config gating that `[[mounts]]`/`[[sockets]]` already have (`coi trust`/`coi untrust`); catalog references carry the same trust level the built-in tool credentials already do, since their host path comes from the catalog rather than the referencing config.

## Test plan

- `go build ./...` and `go test ./...` pass (Linux; this repo's `internal/monitor` needs a Linux host).
- `TestSetupCredentials_Integration` and `TestLoad_CredentialsSurviveRealPath` exercise the real paths end to end: an actual container push/chown/chmod, and a real TOML file loaded through schema validation, merge, and untrusted-marking.
- Manually: a profile with `[[credentials]] bundle = "ollama"` loads via `coi profile show` and seeds `~/.ollama/id_ed25519` into the container on `coi shell`.

Closes #549.


docs/superpowers/plans/2026-07-08-credential-catalog.md
technicalpickles, line 1: I'm not sure what your preference is for having plans like this checked in. I don't see any existing ones though.

mensfeld:
Pushed `482ecb0` (maintainer edit) to address the one duplication concern from review: the copy *values* were already single-sourced through the catalog, but the copy *mechanics* existed twice — `setupCredentials` (mkdir/push/chown/chmod) and `setupCLIConfig`'s essential-file loop — with the catalog comment asking humans to keep them in sync.

The commit extracts `session.seedHostFile` as the single seeding primitive (stat/skip-if-missing → mkdir -p → push → chown-to-code-user → optional chmod) and routes both paths through it. The `[[credentials]]` path is behavior-identical (your incus-gated integration test covers it unchanged); the builtin-tool path picks up two benign deltas noted in the commit message — per-file chown as files land (strictly safer if setup aborts before the recursive chown) and full-path log lines. Catalog comment updated to reflect that there's no longer anything to manually keep in sync.

Rest of the PR looks good — the mount/socket/credential split is the right decomposition (sockets for brokered secrets, mounts for live shares, credentials for copy-then-own files), and reusing the existing trust fingerprint machinery instead of a parallel gate was the right call.

Issue #588, opened 2026-07-10
coi tmux capture/send/list exec as root, fail against the code-user tmux session


technicalpickles:
## Summary

`coi tmux capture`, `coi tmux send`, and `coi tmux list` exec into the container as **root**, but the tmux session created by `coi shell --background` runs as the **`code`** user. Since tmux sessions are per-user (`/tmp/tmux-<uid>/default`), the root-run commands connect to the wrong socket and fail — or, for `list`, silently report no sessions even when one is running.

## Steps to reproduce

```console
$ coi shell --background
...
Created background tmux session: coi-coi-b4487053-1
Use 'coi tmux capture coi-b4487053-1' to view output
Use 'coi tmux send coi-b4487053-1 "<command>"' to send commands

$ coi tmux capture coi-b4487053-1
Error: failed to capture tmux output: exit status 1: error connecting to /tmp/tmux-0/default (No such file or directory)

$ coi tmux list
No active sessions
```

Confirmed the session really is running, just under a different user:

```console
$ incus exec coi-b4487053-1 -- tmux ls
error connecting to /tmp/tmux-0/default (No such file or directory)

$ incus exec coi-b4487053-1 -- su - code -c "tmux ls"
coi-coi-b4487053-1: 1 windows (created ...)
```

`incus exec --user <code-uid> --group <code-uid> -- tmux capture-pane -t coi-coi-b4487053-1 -p` works correctly.

## Root cause

In `internal/cli/tmux.go`, all three commands build a `container.ExecCommandOptions` without setting `User`/`Group`, so `Manager.ExecCommand` (`internal/container/manager.go`) omits `--user`/`--group` from the `incus exec` invocation and defaults to root:

- `tmuxSendCommand` — `internal/cli/tmux.go:76`
- `tmuxCaptureCommand` — `internal/cli/tmux.go:108`
- `tmuxListCommand` (the `tmux has-session` check) — `internal/cli/tmux.go:153`

Compare with `internal/cli/container.go:193` / `:264`, which do set `opts.User` for other exec paths.

The container's `code` user is not always UID 1000 (`container.CodeUID`) — e.g. on hosts where the host UID differs from 1000, `ConfigureUIDMapping` (`internal/session/setup_helpers.go`) applies `raw.idmap` and remaps `code` to the host's UID instead. So the fix needs to resolve the *effective* code UID for the target container (however the shell/run pipeline already does it) rather than hardcoding `container.CodeUID`.

## Suggested fix

[...]

PR #591, opened 2026-07-11
feat(session,tmux,attach): record session exec UID as user.coi.uid metadata (#588 follow-up)

mensfeld:
Implements the follow-up noted in #590's review discussion: the session's exec UID becomes recorded container state instead of something every consumer re-derives.

## What changes

**One authority for the tmux socket.** Session setup resolves `result.ExecUID` and records it as `user.coi.uid` container config (the `user.coi.alias` pattern). The session creation itself (`buildContainerEnv`) execs as that value, and the consumers (`coi tmux capture/send/list`, `coi attach`) read the same metadata via `session.EffectiveCodeUID` — so producer and consumers agree **by construction**, not by two resolutions that happen to match.

This structurally fixes the two corner cases documented in #590:
- **Reuse after a `code_uid` change**: on reuse, setup now trusts the container's recorded/actual state over the new config value (no remap runs on reuse, so the old behavior created the session on a socket the container's user never had — the helpers' probe then disagreed).
- **Cross-project `code_user` names**: metadata carries a number; no user name is involved when reading it, so a container created under another project's custom `[incus] code_user` resolves correctly from any directory.

**Cheaper helpers.** `coi tmux list` drops from 3 incus round-trips per container (`Running()` + `id -u` probe + `has-session`) to **1 + `has-session`**: new `container.ListContainersInfo` surfaces name, running state, and the recorded UID from the single `incus list --format=json` call that `ListContainers` already made (and now delegates to). `send`/`capture`/`attach` replace their exec probe with a plain config read. The value is readable even on stopped containers.

**Backward compatible.** Pre-metadata containers fall back to the live probe (unchanged #590 semantics, including the stderr-based error classification) and are lazily upgraded on first use. The existing tmux e2e tests exercise exactly this fallback path, since `coi container launch` writes no metadata.

## Tests

- Unit: resolution order pinned (metadata hit **never** probes — the fake errors if consulted; miss → probe; probe errors propagate, never silent root), plus metadata parsing (unset/garbage/read-error → fallback).
- E2E (new, discriminating): session created as uid **501** with `user.coi.uid=501` recorded but passwd still mapping `code` to **1000** — a live probe would answer 1000 and target the wrong socket, so the test passes **only** via the metadata path. Existing default-uid and remapped-uid tests keep covering the probe fallback and the no-root-socket invariant.

Changelog under 0.10.1 Improvements.


mensfeld:
Closing unmerged after an adversarial review of this PR found permanent-poisoning modes in the recorded-authority design: a transient probe failure at fresh launch would durably record uid 0 (converting a persistent container's sessions to root with no self-healing, since existing metadata blocks re-probing); the consumer-side lazy write let a read-only `coi tmux list` from a differently-configured project permanently rewrite another container's recorded UID, which session setup then trusted for session creation; and the ownership half of setup (socket proxy uid, chowns of restored state/credentials) was still on the config authority, so the intended divergence path produced sessions that couldn't read their own credentials or ssh agent socket. Full findings are in the review record.

The cost/benefit doesn't hold up: the motivating corner case (cross-project custom `code_user` names) is an exotic, opt-in configuration with zero user reports and is already documented as a limitation, and the remaining benefit was ~100-300ms per helper call — not worth caching *which user your sessions run as* in state with no invalidation. Master with #590 probes fresh on every call: marginally slower, stateless, and self-healing.

If tmux-helper latency ever matters in practice, the safe salvage is the batching-only piece (reuse the single `incus list` JSON for status in `tmux list`, no metadata involved).

PR #597, opened 2026-07-12
fix(shell): close/poweroff cleanup — wait for guest shutdown, truthful messages

mensfeld:
`close` (and all power wrappers) shuts the guest down, so session cleanup must always end with a **stopped** container — deleted if ephemeral, kept-but-stopped if persistent — and say so. Two defects broke that contract (reported in review of a real session transcript):

1. **Persistent branch lies unconditionally** (`cleanup.go`): it prints *'Container kept running - use coi attach to reconnect...'* whenever the container merely EXISTS — including right after `close` powered it off.
2. **Ephemeral 5s race**: cleanup polls `Running()` for a fixed ~5s to distinguish poweroff from a normal `exit`; a guest shutdown slower than that is mislabeled as a user exit and the ephemeral container leaks instead of being removed.

## Two-phase proof

- **Commit 1 (this push): tests only — shell-ephemeral and shell-persistent lanes are expected RED.**
  - persistent: `close` → container must be kept AND reach Stopped, and cleanup must not claim it was 'kept running'
  - ephemeral: guest pinned in 'stopping' for ~30s (slow-ExecStop transient unit + non-forced poweroff; the stock wrappers' `--force` skips unit stops, which is why existing poweroff tests always fit the window) → container must still be recognized as powered off and deleted
- **Commit 2 (follows): the fix** — probe the guest (`systemctl is-system-running` → 'stopping') to detect an in-progress shutdown explicitly instead of inferring from a fixed window; wait up to `[container] shutdown_timeout` for the stop, then delete (ephemeral) or keep and report stopped (persistent). Normal `exit` keeps returning immediately.

Do not merge until phase 2 is pushed and green.


mensfeld:
## Two-phase proof recorded

**Phase 1 (tests-only, run [29191436211](https://github.com/mensfeld/code-on-incus/actions/runs/29191436211) on #598):**
- `shell-persistent` **FAILED as expected** — the log shows the exact reported symptom: `[cleanup] Container kept running - use 'coi attach' to reconnect...` printed right after `close` powered the container off. This is the reported transcript reproduced verbatim in CI.
- `shell-ephemeral` passed on master: with this test's timing, the session-save step hit its internal mid-shutdown handling (wait for full stop, retry via direct file access), so the subsequent state poll saw a stopped container and deleted it. The 5s-window race is only reachable when the save completes while the guest is still stopping (larger tool config dirs, slower pulls) — the test therefore stands as **contract coverage** (deletion after a slow shutdown must hold on every internal path) rather than a red proof.

**Phase 2 (fix, run [29191106278](https://github.com/mensfeld/code-on-incus/actions/runs/29191106278)): fully green** — both new tests plus the whole suite.

Closing #598 (proof recorded).

PR #608, opened 2026-07-16
fix(remap): don't abort UID remap when a read-only mount lives under /home/code

technicalpickles:
## Background

Running `coi shell` in a workspace produced a wall of `Read-only file system` errors during setup and aborted before the container was usable:

```
[setup] Remapping user code from UID 1000 to 501...
Error: setup-container: failed to setup session: failed to remap user code to UID 501: exit status 1: chown: changing ownership of '/home/code/.claude/plugins/marketplaces/claude-plugins-official/.git/branches': Read-only file system
chown: changing ownership of '/home/code/.claude/plugins/marketplaces/claude-plugins-official/.git/hooks/applypatch-msg.sample': Read-only file system
chown: changing ownership of '/home/code/.claude/plugins/marketplaces/claude-plugins-official/.git/hooks/commit-msg.sample': Read-only file system
... (dozens more lines, one per file under the mounted marketplace repo)
```

The triggering config (global `~/.coi/config.toml`) sets a non-default `code_uid` (to match the host UID) and declares a read-only mount under `/home/code`:

```toml
[incus]
code_uid = 501

# Share Claude Code plugins (marketplaces, installed plugin cache/config)
# read-only into every container, so plugins installed on the host are
# available without reinstalling per-session.
[[mounts.default]]
host = "~/.claude/plugins"
container = "/home/code/.claude/plugins"
readonly = true
```

Neither piece is coi built-in behavior — both are user-declared config. `code_uid` differing from the image's default 1000 makes coi remap the `code` user's UID/GID at setup time; the read-only `[[mounts.default]]` entry is just a normal user mount that happens to land under the code user's home directory.

The interaction: `usermod -u`, as a side effect of changing the UID, walks the whole home directory chowning files it finds owned by the old UID — even without `-m`. That walk hit the read-only-mounted `.claude/plugins` tree and made `usermod` exit non-zero, even though the actual UID/GID change had already applied. coi treated that non-zero exit as fatal and aborted the whole session setup. Any read-only mount under `/home/code` combined with a `code_uid` remap triggers the same failure — the plugins mount above is just the config that surfaced it.

Notably, coi already had a "best-effort" tolerance for exactly this failure mode on the explicit follow-up `chown -R /home/code` (see the existing comment referencing #534) — it just wasn't applied to the `usermod` call itself, which is where this particular case actually failed.

## Summary
- `usermod -u` walks the `code` user's home directory to fix up file ownership after a UID change (even without `-m`). When a read-only mount already sits under `/home/code` at that point, that walk hits `Read-only file system` and `usermod` exits non-zero even though the UID/GID change itself succeeded — this was aborting `coi shell`/`coi run` setup entirely instead of just warning, the way the existing `chown -R` fallback already does.
- Verify the UID actually landed (via a follow-up `id -u` probe) before treating a non-zero exit as fatal, in both places this logic is duplicated: `internal/cli/run.go` (`coi run`) and `internal/session/setup.go` (`coi shell`).

## Test plan
- [x] Reproduced the original failure end-to-end (`coi shell` with `code_uid` remap + a read-only mount under `/home/code`) and confirmed it now completes setup, with `id code` reporting the correct remapped UID/GID inside the container.
- [x] Added fast, deterministic unit tests (`internal/cli/remap_test.go`) covering: tolerating the read-only-mount chown failure, still failing hard on a genuine remap failure, and the no-op path when no remap is needed.
- [x] `go build ./...` and `go vet ./...` clean.
- [x] Full `internal/cli` test suite passes.

https://claude.ai/code/session_01KfUhvvD2MGKD61EKJ8oAf1


mensfeld:
Reviewed — the diagnosis is correct and I verified the load-bearing claim empirically (`usermod -u` with a read-only mount under the home: exits `12`/`E_HOMEDIR` with the passwd update already committed, so the UID probe is a sound discriminator; since `groupmod` runs first and `usermod` writes uid+gid in one passwd update, a confirmed UID implies the full remap landed). Nice find, and thanks for covering both duplicated call sites + the unit tests.

Pushed two commits on top (maintainer edit):

1. **Merge of current master** (post #610/#615) so CI runs against the latest code.
2. **`43d91b1` — closes the one gap the review found + adds e2e coverage:**
   - The `coi shell` chain fused `groupmod && usermod && chown -R` into one command, so when `usermod` itself exits `E_HOMEDIR` mid-walk (the failure mode this PR tolerates), the `&&` chain skipped the trailing `chown -R` — the warning path then proceeded with writable home files still owned by UID 1000, leaving the remapped user unable to write its own dotfiles. Restructured to match `coi run`: `groupmod && usermod` fatal-with-probe (your logic unchanged), `chown -R` separate and always best-effort.
   - Added an e2e regression test (`tests/shell/ephemeral/remap_with_readonly_home_mount_ephemeral.py`): `coi shell` with `code_uid = 1001` + a read-only `[[mounts.default]]` under `/home/code` containing an old-UID-owned file — deterministically fatal pre-fix, exercises both failure modes, and asserts the session comes up with `id -u code` reporting the remapped UID (so the tolerance can't mask a remap that never applied).
   - CHANGELOG entry.

Will merge once CI is green.

PR #617, opened 2026-07-18
Review fixes for #603: H1 metadata-ordering, permanent DNS elements, live-set monitor, auto-merge

mensfeld:
Draft fixes from a code review of #603 (allowlist DNS enforcement), targeted at that branch so the delta is reviewable on its own. Four independent commits — accept, drop, or reshape any of them.

The core design of #603 held up under review: the egress boundary is the nft set + default-reject (a bypassed DNS still can't reach a non-set address), the two-chain DNS block ordering is correct, the `AllowDynamicIPs` lock is genuinely held across the exec, the removed DNS proxy left no redirect leak, and setup fails closed. These four address the gaps found around it.

## H1 — reject private/metadata ranges BEFORE the set-accepts `(2a638e8)`
🔴 The per-set accept rules were appended ahead of the RFC1918 / link-local reject, and resolved answers enter the set unfiltered. An allowlisted hostname that resolves (on the host) to `169.254.169.254` — DNS rebinding on an attacker-influenced/compromised allowlisted domain, or a poisoned record — was matched by the set-accept before the reject ran, reaching the **cloud metadata endpoint (credential theft)** or internal hosts. Pure reorder into one accept-or-reject before the set-accepts (mirrors restricted mode); **semantics unchanged**. Validated against the real kernel — the DNS-block integration test gains an assertion that `169.254.0.0/16` precedes the set-accept.

## #2 — DNS-learned dynamic elements are now permanent `(08c22b8)`
🔴 A finite kernel timeout is only safe if the in-process refresher renews it, but the refresher can't outlive the container: `coi shell --background` returns in seconds (deferred Teardown cancels it) and tmux-detach exits the process — the deployments the design advertises as surviving a `coi` exit. A finite timeout evicted a still-valid address ~20 min in with nothing to renew it → the container resolved an allowlisted name from `/etc/hosts` to an address the firewall no longer held → default reject → "Unable to connect", non-self-healing (the exact failure this mode prevents). Same reasoning that removed the DNS proxy. Tradeoff noted in-code: bounding set growth should move to refresher-driven pruning (follow-up), not a kernel timeout.

## M1 — the monitor reads the LIVE nft set `(08c22b8)`
🟡 The monitor re-resolved `allowed_domains` independently into a frozen snapshot, so for a rotating frontend it drifted from the firewall's evolving set; a legit connection the firewall permitted looked unauthorized → under `auto_pause_on_high`, paused the container mid-task. New `CurrentAllowlistCIDRs()` reads the static+dynamic sets (parser unit-tested against real nftables JSON); the collector gains a fail-safe `AllowedCIDRsProvider`, wired in allowlist mode, with the static snapshot as fallback.

## M2 — `auto-merge` on the static set `(08c22b8)`
🟡 Without it, overlapping/duplicate literal entries (a CIDR + an address inside it; two overlapping published ranges; `8.8.8.8` + `8.8.8.8/32`) make nft reject "conflicting intervals", `ApplyAllowlist` fails, and — fail-closed — the container launches with no egress at all. Verified against nftables v1.0.9.

## Validation
`go build`, `go vet`, `golangci-lint` (0 issues), `gofmt`, and network/monitor/cli tests all pass, plus new unit tests (nft-set parser, collector fallback) and the H1 real-nft integration assertion. The live e2e for #2/M1 (real container + rotating frontend + auto-pause) is the `network` CI lane's job.

## Deliberately NOT changed (author's call)
- Allowlist mode still keys private/metadata blocking on `allow_local_network_access` alone, not `block_private_networks`/`block_metadata_endpoint` separately (so metadata is still reachable if local access is on) — a semantics change, left as a follow-up.
- #2 removes kernel-timeout eviction; refresher-driven pruning to bound set growth is a follow-up.

Full review notes (incl. the lower-severity items — IPv6 input, DoH, kill idempotency, hosts atomicity) available on request.


mensfeld:
Folded into #603 directly. The two fix commits here (H1 — reject RFC1918/metadata before the allowlist set-accepts; and the #2/M1/M2 fixes — permanent DNS elements, live-set monitor provider, auto-merge on the static set) were rebased on top of the `fix/allowlist-dns-enforcement` branch along with current `master` (#602/#608/#610/#615/#616). #603 is now mergeable and its CI is running. Closing this as its content now lives in #603.

PR #650, opened 2026-07-27
Fix skipDangerousModePermissionPrompt nesting so bypass-mode warning is suppressed

### LINKED ISSUE #649: ClaudeTool sandbox settings nest skipDangerousModePermissionPrompt under permissions, but Claude Code reads it top-level
`ClaudeTool.GetSandboxSettings()` writes `skipDangerousModePermissionPrompt` into the nested `permissions` object it injects into `settings.json`/`.claude.json`:

https://github.com/mensfeld/code-on-incus/blob/master/internal/tool/tool.go#L145-L174

```go
settings["permissions"] = map[string]interface{}{
    "defaultMode":                       "bypassPermissions",
    "skipDangerousModePermissionPrompt": true,
}
```

Claude Code doesn't read it from there. It only suppresses the "Bypass Permissions mode" warning banner when `skipDangerousModePermissionPrompt` is a **top-level** key in `settings.json`, alongside `allowDangerouslySkipPermissions` and `bypassPermissionsModeAccepted` (which *are* already top-level and do work).

Net effect: every `coi shell` session with the default (non-`interactive`) permission mode still shows this on first launch of a fresh container, even though the injected settings already have `bypassPermissionsModeAccepted: true`, `allowDangerouslySkipPermissions: true`, and `permissions.defaultMode: bypassPermissions`:

```
WARNING: Claude Code running in Bypass Permissions mode
...
❯ 1. No, exit
  2. Yes, I accept
```

Requires one manual keypress per fresh container to get past, which defeats the point of `coi` pre-accepting bypass mode on the user's behalf.

### Repro

1. `coi build` (or any image with the default Claude tool)
2. `coi shell` in a fresh workspace
3. Confirm `settings.json` in the container has `permissions.skipDangerousModePermissionPrompt: true` (from the nested injection above) - warning still shows.
4. Manually add `skipDangerousModePermissionPrompt: true` as a **top-level** key in the container's (or host source) `settings.json`, relaunch - warning is gone, session goes straight to the REPL.

### Suggested fix

Move `skipDangerousModePermissionPrompt` out of the nested `permissions` map to a top-level key in the map `GetSandboxSettings()` returns, e.g.:

```go
settings["allowDangerouslySkipPermissions"] = true
settings["bypassPermissionsModeAccepted"] = true
settings["skipDangerousModePermissionPrompt"] = true
settings["permissions"] = map[string]interface{}{
    "defaultMode": "bypassPermissions",
}
```

### Environment

- `coi version latest-12-gf0e1098`
- Claude Code v2.1.215
- Incus containers on an OrbStack Ubuntu VM

technicalpickles:
## Summary

- `ClaudeTool.GetSandboxSettings()` nested `skipDangerousModePermissionPrompt` under the `permissions` map, but Claude Code only reads it as a top-level `settings.json` key
- Every `coi shell` session in bypass mode (the default) still showed the "Bypass Permissions mode" warning banner on first launch of a fresh container, requiring a manual keypress despite `allowDangerouslySkipPermissions` and `bypassPermissionsModeAccepted` already being set correctly
- Moves the key to the top level alongside those two, verified end-to-end with a real `coi shell` session landing straight in the REPL with no prompts

Fixes #649.

## Test plan

- `go test ./internal/tool/... ./internal/session/...`
- Manually confirmed with a throwaway `coi shell --background` session


mensfeld:
Reviewed and pushed one maintainer commit to this branch.

**Code review — the fix is correct. ✅**
`skipDangerousModePermissionPrompt` was nested under `permissions`, but Claude Code only reads it at the **top level** of `settings.json`, so the bypass-mode warning wasn't suppressed (#649). Hoisting it to top-level — alongside the sibling `allowDangerouslySkipPermissions` / `bypassPermissionsModeAccepted` flags that were already top-level — is the right fix. All three test files were updated to assert the new placement (top-level present, no longer nested), and the `internal/tool` + `internal/session` suites pass locally.

**Removed the `docs/superpowers/` files (maintainer commit).**
This branch carried two unrelated commits ("Add design spec / implementation plan for generic credential catalog (#549)") that added `docs/superpowers/plans/…` and `docs/superpowers/specs/…` (~2.3k lines). Those are superpowers plan/spec tooling scratch, not product docs — and the in-repo `docs/` tree was just removed in #652 (documentation now lives in the [GitHub wiki](https://github.com/mensfeld/code-on-incus/wiki), and `docs/` is gitignored). I dropped them here so this PR is scoped to just the `skipDangerousModePermissionPrompt` fix. No change to your code.

mensfeld (approved):

PR #655, opened 2026-07-27
test(monitoring): stabilize flaky connection-detection + log-rotation tests

mensfeld:
## Problem

`test_network_connection_detection.py::test_suspicious_tcp_port_detected` flaked on a loaded `monitoring-network-detection` lane (surfaced on #650's CI, unrelated to that PR):

```
AssertionError: TCP connection to :4444 never reached ESTABLISHED after 3 attempts —
the in-container connect() failed every time (server or routing problem)
  Port 4444 listening: True
```

The server **was** listening — the *connect* is what failed.

## Root cause

The test makes the container connect to its **own routable IP** on port 4444 (a hairpin route) and waits for `ESTABLISHED` so the monitor can observe it. The connect was a single fire-and-forget one-liner with **no timeout**:

```python
s.connect((container_ip, 4444)); time.sleep(120)
```

If that one `connect()` raised — a dropped SYN on the hairpin, an `http.server` still warming up, or slow `python3` startup under CI load — the process died immediately with **no ESTABLISHED socket**. The outer loop retried only 3×, and each failed attempt then wasted ~20s in `_wait_port_established` polling for a socket that would never appear. Under load, all three attempts lost.

This is a **known-flaky test**, previously patched several times (`bcd13d7` "fix flaky TCP connection-detection test (#615)", `91bd859` widen poll window 40s→75s, `51b55c5` wait for listener). It is **green on master** — intermittent, not a regression.

## Fix

Make the connect **self-retry inside the container** so a transient failure re-attempts in ~2s instead of forfeiting a whole outer attempt:

```python
sock = None
deadline = time.time() + 120
while sock is None and time.time() < deadline:
    try:
        sock = socket.create_connection((container_ip, 4444), timeout=5)
    except OSError:
        time.sleep(2)
if sock is not None:
    time.sleep(300)
```

Plus:
- invoke `python3 -c` **directly** (drop the `bash -c` wrapper — no quoting/heredoc pitfalls);
- widen the per-launch establish-wait 20s → **90s** (the connect keeps retrying, so give it room before re-launching);
- keep the 3× outer loop purely as a backstop for a dead `incus exec` session.

**Doesn't mask a real bug:** if the connection establishes but no threat is produced, Phase 2 still fails red (a genuine detector miss). Only the transient *connect* failure is now tolerated.

Verified `py_compile` + `ruff check`/`format` clean, and the embedded connect script compiles as valid Python. No product code change; test-only.


mensfeld:
**Folded in a second stabilization + re-triggered CI.**

A different, unrelated lane (`monitoring-log-inotify`) flaked on the prior run — `test_log_rotation_handled` reported `line_present_in_container=True, events: []` (the auth line landed but produced no threat), even on a lightly-loaded lane (3 tests, 144s).

**Root cause:** `auth.log` doesn't exist when the monitoring daemon starts, so first detection depended on catching the file's *creation* via the parent-directory `IN_CREATE` watch — a path `logwatcher.go` documents as unreliable across the overlayfs namespace boundary.

**Fix:** pre-create an empty `auth.log` during the settle window so the watcher registers a **direct file watch**, making the trigger write a plain append/`IN_MODIFY` on an already-watched file (the reliable path). Also widened the settle wait (3s→5s) and both write-detect budgets (60→90 polls). A genuine detector miss still fails red.

Both stabilizations are now in this PR, so the re-triggered run exercises both fixes together.

PR #662, opened 2026-07-27
fix(install): fall back to btrfs when ZFS storage is unavailable (#661)

### LINKED ISSUE #661: install.sh's ZFS storage setup can't work on OrbStack (no kernel module possible); btrfs works and gets a similar speedup
Ran into this setting up an OrbStack VM as a coi host. `setup_zfs_storage()` in `install.sh` tries to give containers fast CoW storage via ZFS, but on OrbStack this can't ever succeed - not a config mistake, just how OrbStack VMs are built.

## Version
- coi install.sh from master (`setup_zfs_storage` around line 488-540 as of this filing)
- Host: OrbStack 2.2.1, Ubuntu 24.04 (noble) VM, guest kernel `7.0.11-orbstack-00360-gc9bc4d96ac70`

## Repro
```
$ sudo apt-get install -y zfsutils-linux
# installs fine - userspace tools/libs only, no kernel module
$ sudo incus storage create zfs-pool zfs size=50GiB
# fails - same underlying cause install.sh's own zpool creation hits
```
The real error is one level down, from modprobe:
```
modprobe: FATAL: Module zfs not found in directory /lib/modules/7.0.11-orbstack-00360-gc9bc4d96ac70
```

## Where it comes from
OrbStack Linux VMs have no kernel packages at all - no `linux-image`, no `linux-headers`, no `/boot`. The kernel is injected by the host, not managed by apt inside the guest. ZFS (OpenZFS) is an out-of-tree module, so it needs either a precompiled module matching the exact kernel build (doesn't exist for OrbStack's custom kernel) or DKMS + headers to build one (no headers, ever). There's no route to a working `zfs.ko` here.

Not specific to my VM or to Ubuntu - orbstack/orbstack#2169 shows the identical `modprobe: FATAL: Module zfs not found` error on a Rocky Linux OrbStack VM. And the general mechanism this would need (loading custom/DKMS kernel modules in an OrbStack VM) has been an open OrbStack feature request since 2023: orbstack/orbstack#258.

`setup_zfs_storage()` degrades semi-gracefully today (best-effort, `|| true` in `install()`), but the failure message ("ZFS storage pool creation failed") doesn't say why, so there's no way to tell "unfixable on this platform" from "fluke, worth retrying."

## btrfs works, and gets most of the way there

btrfs is compiled directly into OrbStack's kernel (no module needed) - OrbStack's own VM root disk is already btrfs. Tested it directly against incus:

```
$ sudo apt-get install -y btrfs-progs && sudo systemctl restart incus
$ incus storage create btrfs-pool btrfs size=50GiB
Storage pool btrfs-pool created
$ incus profile device set default root pool=btrfs-pool
```

Timed container launches on it: first launch (cold image pull) 8.0s; second launch, same cached image, 0.167s. Same cached image on the existing `dir` pool: 1.14s. Roughly 7x faster than `dir` - short of ZFS's claimed ~50ms elsewhere, but a real, working win on a platform where ZFS isn't an option at all.

## Suggestion

Detect OrbStack specifically (coi's own `internal/vmhost/vmhost.go` already does this exact check - guest kernel `osrelease` containing `"orbstack"` - for a different purpose) and skip the ZFS attempt with a clear "not supported on OrbStack" message, offering btrfs instead. Happy to send a PR for a `setup_btrfs_storage()` alongside (or in place of) the zfs one if that's a direction you'd want.

technicalpickles:
Fixes #661.

## Problem

`setup_zfs_storage()` was the installer's only fast-storage path. On a host where the ZFS kernel module can't load, it printed a bare `⚠ ZFS storage pool creation failed` and containers silently stayed on the slow default `dir` pool. The message gave no way to tell "unfixable on this platform" from "fluke, worth retrying."

OrbStack is the unfixable case: the guest gets its kernel from the host, with no kernel packages, no headers and no module tree, so the out-of-tree ZFS module can neither ship precompiled nor be built with DKMS. There is no route to a working `zfs.ko`. Details and repro in #661.

## Change

`setup_zfs_storage` becomes one arm of a new `setup_fast_storage` chain:

- **`is_orbstack()`** matches `*orbstack*` in `uname -r`, the same signal `internal/vmhost/vmhost.go` already keys off. On OrbStack the ZFS attempt is skipped entirely, with a message saying why, and btrfs is set up instead.
- **Everywhere else**, ZFS is tried first, unchanged. If the pool can't be created, the underlying error is printed (so you see the actual `modprobe: FATAL: Module zfs not found` instead of just "failed") and it falls back to btrfs.
- **`setup_btrfs_storage()`** mirrors the ZFS one: installs `btrfs-progs` (`btrfsprogs` on zypper) if missing, restarts Incus so it re-probes its storage drivers (Incus won't see a freshly installed `mkfs.btrfs` otherwise), creates `btrfs-pool` at 50GiB, and repoints the `default` profile's root device.
- The `Containers will use default storage (slower but functional)` warning moved to the end of the chain, so it only fires when both drivers fail.

btrfs is still copy-on-write, so it gets most of the way there. Measured on an OrbStack Ubuntu 24.04 guest, launching from a cached image: **0.167s on `btrfs-pool` vs 1.14s on `dir`**, roughly 7x. Short of ZFS's ~50ms, but a real win on a platform where ZFS isn't an option at all.

**On a host where ZFS works, behavior is unchanged.**

## Tests

Four new cases in `internal/installer/install_sh_test.go`, in the existing stub-the-binaries-on-PATH style:

- `TestInstallSh_IsOrbstack_DetectsKernelRelease` — orbstack, mixed-case orbstack, stock Ubuntu kernel
- `TestInstallSh_SetupFastStorage_SkipsZfsOnOrbStack` — asserts ZFS is never attempted and btrfs is set up
- `TestInstallSh_SetupFastStorage_FallsBackToBtrfsWhenZfsFails` — asserts the ZFS failure reason is surfaced, btrfs takes over, and the default-storage warning does not appear
- `TestInstallSh_SetupBtrfsStorage_SkipsWhenPoolExists` — idempotence

The existing `TestInstallSh_SetupZfsStorage_SuppressesOutputOnSuccess` still passes unmodified.


mensfeld:
Reviewed this in depth (checked out the branch, read all three storage functions in full context, ran the tests + shellcheck, and verified the cross-references). **It's a solid, well-tested PR — no correctness bugs.** Verified: `pkg_install`'s 4-arg order makes `btrfs-progs … btrfsprogs` correct per distro; `set -euo pipefail` doesn't bite (the chain runs in condition/`|| true` contexts where `-e` is suppressed); the OrbStack `uname -r` match is identical to `vmhost.go:60`'s `osrelease` check; the removed "default storage" warning correctly relocated to fire only when the whole chain fails; no orphaned callers; shellcheck clean on the new functions.

I pushed a maintainer commit addressing the one substantive finding + adding integration tests:

**1. Daemon-readiness after restart (finding #1).** `setup_btrfs_storage` restarts Incus to pick up the freshly installed driver, then immediately issues `incus storage list`/`create` — which can race the API socket coming back up (`systemctl restart` returns when the unit is *active*, a beat before the socket is ready). Extracted a **`restart_incus()`** helper = `systemctl restart` + **`incus admin waitready --timeout=30`** (bounded, best-effort), and used it at **both** restart sites — the new btrfs-driver probe *and* the pre-existing idmap re-read at the top of the file, which had the same latent race.

**2. Integration tests** (real bash, stubbed binaries, same style as the existing suite):
- `TestInstallSh_RestartIncus_WaitsReadyAfterRestart` — `waitready` runs *after* the restart.
- `TestInstallSh_SetupBtrfsStorage_WaitsReadyBeforeCreatingPool` — drives the fresh-install path (forces the install branch by dropping `/usr/sbin` from PATH) and asserts the ordering **restart → waitready → pool-create**.

**3. CHANGELOG** — resolved the conflict from master (kept both entries) and noted the waitready hardening in the btrfs entry.

All 17 installer tests pass, `go build ./...` green, shellcheck clean on the new code. Three remaining review notes are optional polish (btrfs≈zfs function duplication; `setup_fast_storage`'s implicit `return 0` on btrfs success; the restart-only-on-fresh-install assumption) — none block merge. Nice fix; the OrbStack rationale and the 7x measurement are convincing.

PR #673, opened 2026-07-30
fix(version): stop double-'v' prefix that broke release builds and `coi update`

sklarsa:
## Problem

On an installed release, `coi version` prints a doubled `v`:

```
code-on-incus (coi) vv0.10.1
```

and `coi update` **wrongly reports "You are already on the latest version"** even though a newer release exists:

```
Current version: vv0.10.1
Latest release:  v0.11.0

You are already on the latest version.
```

This affects every downloaded release binary (what `install.sh` fetches). Locally-built binaries (`make build` with no `VERSION` in env) are unaffected.

## Root cause

The release workflow feeds the raw git tag (e.g. `v0.10.1`) to `make build` through the environment (`release.yml` → `version=${GITHUB_REF#refs/tags/}`, which only strips `refs/tags/`, keeping the `v`).

The Makefile normalizes with:

```make
VERSION ?= $(shell (git describe --tags --always --dirty ... ) | sed 's/^v//')
```

But `?=` only assigns when unset. Since the workflow already exports `VERSION`, the `sed 's/^v//'` **never runs**, so the binary is built with `cli.Version="v0.10.1"`. The print paths then do `v%s` → `vv0.10.1`.

Worse than cosmetics: `compareVersions` splits on `.`, so `"v0"` fails `strconv.Atoi` and falls back to string comparison, where `"v0" > "0"`. The current version therefore looks `>=` latest and the update is suppressed — exactly the reported symptom.

The existing CI "Verify version" guard missed it because `grep -oP 'v\K[0-9]+\.[0-9]+\.[0-9]+'` matches the digits *inside* `vv0.10.1` and extracts `0.10.1`, passing.

## Fix (defense-in-depth)

1. **Workflow boundary** — strip the leading `v` before `make build`, honoring the Makefile's bare-version contract.
2. **CLI robustness** — add `normalizeVersion()` and apply it wherever `Version`/tags are read (`version` text + JSON output, `update` current + latest), so a stray `v` can never reach display or comparison again.
3. **Tighten the verify guard** — compare the full version line against the exact expected string instead of a `\K` substring, so any doubled/missing `v` fails the release.

## Tests

- `TestNormalizeVersion` — bare/`v`/`vv`/`dev`/empty inputs.
- `TestCompareVersionsToleratesVPrefix` — reproduces the suppressed-update case (`vv0.10.1` vs `v0.11.0` → update available).

Verified locally that a binary built with the buggy `Version=v0.10.1` now prints `v0.10.1` and reports the update correctly.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -87,6 +87,11 @@ jobs:
           CC: ${{ matrix.arch == 'arm64' && 'aarch64-linux-gnu-gcc' || 'gcc' }}
           PKG_CONFIG_PATH: ${{ matrix.arch == 'arm64' && '/usr/lib/aarch64-linux-gnu/pkgconfig' || '' }}
         run: |
+          # The Makefile's VERSION uses `?=`, so an env value wins and its own
+          # `sed 's/^v//'` never runs. Strip the leading 'v' here to honor the
+          # Makefile's bare-version contract; otherwise the binary reports
+          # "vv0.10.1" and version comparison breaks (#672 follow-up).
+          export VERSION="${VERSION#v}"
           make build
           mv coi coi-${{ matrix.os }}-${{ matrix.arch }}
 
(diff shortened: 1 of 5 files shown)


mensfeld:
Thanks for this — I reproduced both symptoms from a `Version=v0.10.1` build (`coi version` → `vv0.10.1`; `compareVersions("v0.10.1","0.11.0")=1` → update wrongly suppressed) and confirmed your fix resolves both. 

I pushed one follow-up commit (`5a8f372`) to your branch: the normalize-everywhere pass covered the `coi version` subcommand but not cobra's **`--version` flag**, which renders from `rootCmd.Version` (the raw ldflags value). On a clean build that printed `coi version 0.11.0` (no `v`, inconsistent), and on a mis-tagged build it would print `coi version vv0.10.1` — the same doubled-`v` this PR kills, on a surface it missed.

The commit normalizes `rootCmd.Version` and adds a shared version template so `coi --version` renders the exact same `code-on-incus (coi) v<semver>` line as the subcommand. Verified with a `Version=vv0.10.1` build — both now print `code-on-incus (coi) v0.10.1`. Added `TestRootVersionFlagOutput` (bare/`v`/`vv` inputs) and `TestRootCmdVersionWiring` (guards the wiring can't be silently reverted; confirmed it fails without the fix).

Two lower-priority notes I left as-is (not blocking): the Makefile's `VERSION ?=` + skipped `sed` is still a footgun for any non-release caller passing `VERSION=vX.Y.Z` (a deeper fix would normalize in the Makefile unconditionally), and dev builds render as `vdev`. Happy to fold either in if you'd like. LGTM once CI is green.

mensfeld (approved):

mensfeld:
:bowing_man: amazing work!

mensfeld:
and sorry for the trouble

PR #675, opened 2026-07-31
fix(session): ~/.claude/CLAUDE.md grows on every session (#674)

### LINKED ISSUE #674: Sandbox user's ~/.claude/CLAUDE.md grows with every session resume
I saw a warning in a coi claude session that /home/code/.claude/CLAUDE.md is over the 40.0k-char limit (108.2k chars) . I asked Claude what was in it and he/it says:

    Now I understand the full picture. Here's the diagnosis:
    Why it's 108k chars: The COI sandbox's environment description block (# COI Sandbox Environment ... # COI Sandbox Context) 
    has been appended 16 times to /home/code/.claude/CLAUDE.md — once per container boot/session — instead of
    replacing the previous copy. All 16 copies are functionally identical.

mensfeld:
Fixes #674.

## Bug

`injectAutoContextFile` (`internal/session/setup_context.go`) writes the COI sandbox context block into the tool's auto-context file (Claude's `~/.claude/CLAUDE.md`) at session setup. When the file already existed it **appended** the block — never checking for its own prior copy. On a **persistent container reused across sessions**, `~/.claude/CLAUDE.md` survives between sessions and setup re-runs each time, so the block was appended again every session and the file grew without bound. The reporter hit **16 identical copies / 108k chars**, past Claude Code's 40k-char limit.

## Reproduction — confirmed red on CI, two layers

**1. Unit layer (Go)** — `internal/session/setup_context_autoctx_test.go` drives the real `injectAutoContextFile` across 3 sessions on one persistent home via a fake `ContainerManager`. Commit 1 (test only) was **red** on run [30611781775](https://github.com/mensfeld/code-on-incus/actions/runs/30611781775) (`Unit Tests`): `found 3 copies (18686 chars)`.

**2. End-to-end layer (Python)** — `tests/shell/persistent/claude_md_no_growth_persistent.py` drives the **real `coi` CLI**: it creates a persistent container, then re-enters the *same* container across sessions (stop + `coi shell --container <name> --background`) so setup re-runs its `~/.claude/CLAUDE.md` injection against the persisted home, and asserts one copy. Validated against pre-fix code on a scratch branch — the `shell-persistent` lane failed with:

```
tests/shell/persistent/claude_md_no_growth_persistent.py::test_claude_md_does_not_grow_across_persistent_sessions FAILED
E  AssertionError: #674: ... after 3 sessions on one persistent container,
   but found 3 copies (18416 chars). ...
E  assert 3 == 1
```
(run [30619117109](https://github.com/mensfeld/code-on-incus/actions/runs/30619117109), `shell-persistent`). The test is left in this PR, where it passes with the fix.

> The e2e test leaves the tool at its default (Claude) so the auto-context path fires — only `ClaudeTool` implements `AutoContextFile()`, so a dummy-tool session never exercises it — and uses `--background` (setup + detach) so no real `claude` binary is needed. The reuse guard (no second slot) confirms the sessions hit the same persisted container.

## Fix

The block is now delimited by `# BEGIN COI Sandbox Context …` / `# END COI Sandbox Context` markers and made **idempotent**: each session reads the file, strips any prior coi-managed block, and writes a single fresh one — **overwriting** (regenerating) the dynamic block while **preserving** any non-managed (user/host-copied) content. This matches how `~/SANDBOX_CONTEXT.md` already regenerates each session, without duplicating.

`TestInjectAutoContextFile_PreservesHostContent` locks in that user/host content is kept alongside exactly one managed block. `go build`, `go vet`, `gofmt`, `golangci-lint`, and `ruff` are clean.

**Note on already-affected files:** this stops the per-session growth going forward. A file that already accumulated many *unmarked* copies from the old code won't be retroactively de-duplicated (the old copies have no markers to identify safely); such a file stops growing and can be cleared once if it's still over the limit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## 0.11.1 (Unreleased)
+
+### Fixed
+
+- **The sandbox context block no longer grows `~/.claude/CLAUDE.md` on every session (#674)** — coi injects its sandbox context into the tool's native auto-context file (Claude's `~/.claude/CLAUDE.md`) at session setup. When the file already existed it **appended** the block without checking for its own prior copy, so on a persistent container reused across sessions the block accumulated one copy per session — a reporter hit 16 identical copies (108k chars), past Claude Code's 40k-char limit. The block is now delimited by `# BEGIN COI Sandbox Context …` / `# END COI Sandbox Context` markers and the injection is **idempotent**: each session strips any prior managed block and writes a single fresh one, overwriting (regenerating) the dynamic block while preserving non-managed user/host content. Marker matching is line-anchored so the delimiters can't be spoofed by content that merely mentions them (e.g. a custom `context_file`). The same pass also **heals files already bloated** by the old code — old unmarked copies (identified by the coi-exclusive `# COI Sandbox Context` separator) are removed on the next session, collapsing an accumulated file back to a single block while keeping any real user content. Covered by a deterministic Go test and an end-to-end `shell-persistent` integration test that reproduces the growth across real sessions.
+
 ## 0.11.0 (2026-07-29)
 
 ### Breaking Changes
--- a/internal/session/setup_context.go
+++ b/internal/session/setup_context.go
@@ -64,30 +64,102 @@ func resolveContextContent(info tool.ContextInfo, customPath string, logger func
 // Markers delimiting the coi-managed sandbox context block inside the tool's
 // auto-load file. The block is rewritten (not appended) every session, so the
 // file never accumulates duplicate copies (#674). Anything outside the markers
-// (e.g. a CLAUDE.md copied from the host) is preserved.
+// (e.g. a CLAUDE.md copied from the host) is preserved. Matching is line-anchored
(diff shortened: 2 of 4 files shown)


mensfeld:
**Update — folded in the two code-review findings** (commit `d733613`):

1. **Marker matching is now line-anchored** — a marker string appearing mid-content (e.g. inside a custom `context_file`) can no longer be mistaken for a real delimiter. New guard: `TestInjectAutoContextFile_ContentContainingMarkerStaysIdempotent`.
2. **Already-bloated files are now healed** — the injection also strips legacy unmarked copies (old `# COI Sandbox Context`-separated blocks, recognized by header + a distinctive body fingerprint), collapsing an accumulated file back to a single block on the next session while preserving genuine user/host content. New guards: `TestInjectAutoContextFile_HealsLegacyDuplicates` / `_HealsLegacyPreservingHostContent`.

This supersedes the earlier "already-affected files aren't de-duplicated" caveat — the reporter's 108k file now heals to one block on the next session. Added a `0.11.1 (Unreleased)` CHANGELOG entry. All new guards were verified to fail without their fix.

PR #676, opened 2026-07-31
test: e2e reproduction for #674 (EXPECTED RED — validation)

mensfeld:
Validation-only PR to confirm the end-to-end Python integration test reproduces #674 on pre-fix code. The `shell-persistent` lane is **expected to fail** here (the CLAUDE.md sandbox block is appended every session). Once confirmed red, the test moves into the fix PR #675 (where it passes with the fix), and this PR is closed. Do not merge.


mensfeld:
Validation complete: the end-to-end test reproduced #674 on pre-fix code here (shell-persistent lane failed with 'found 3 copies', run 30619117109). The test has been moved into the fix PR #675, where it passes with the fix. Closing this validation PR.

PR #679, opened 2026-08-02
fix(uidmap): auto-fall back to raw.idmap when idmapped mounts fail on OrbStack (#678)

### LINKED ISSUE #678: vmhost.go assumes OrbStack's virtiofs mounts support idmapped shift, but mine doesn't and there's no fallback
## What's going on

`internal/vmhost/vmhost.go` treats OrbStack as a VM host whose virtiofs mounts are known-good for Incus's idmapped shift mounts:

```go
// OrbStack also mounts the host filesystem via virtiofs, so it must be
// excluded before the generic virtiofs check below, or it would
// false-positive as Lima/Colima. Unlike them, OrbStack's virtiofs mounts
// support idmapped shift mounts fine, detected via its guest kernel's
// release string (e.g. "7.0.11-orbstack-...").
if strings.Contains(strings.ToLower(osRelease), "orbstack") {
    return KindOrbStack
}
```

`KindOrbStack.HandlesUIDMapping()` returns `false` (correctly — OrbStack doesn't map UIDs at the VM level the way Lima/Colima do), so unless `disable_shift` is set manually, coi sends `shift=true` on the workspace device mount (`internal/container/manager.go` `MountDisk`) by default. On my OrbStack host, that assumption is false: the shift mount fails outright, every time, with no automatic fallback.

## Repro

```
$ coi run -- echo hi
Error: Failed to setup device mount "workspace": idmapping abilities are required but aren't supported on system
Warning: UID namespace isolation not supported in this environment, retrying
Error: Failed to setup device mount "workspace": idmapping abilities are required but aren't supported on system
Error: launch-container: failed to launch container: exit status 1
```

Reproduced identically against both a plain `dir` storage pool and a `btrfs` pool, so it's not pool-specific. The "retrying" line is the existing container-start isolation fallback (`security.idmap.isolated`, from #534) — it doesn't help here because the failure is at the *device mount* step, a different code path than what that fallback covers.

## Root cause

The `strings.Contains(osRelease, "orbstack")` branch hardcodes an assumption about guest kernel capability from a string match on the OrbStack marker, not from probing whether idmapped mounts actually work. Whatever OrbStack version/kernel combination the comment's claim held for, it doesn't hold universally — Incus's shift-mount path needs kernel idmap support the guest doesn't have here.

Manually setting `[incus] disable_shift = true` (an existing, documented config field) works around it completely — `ConfigureUIDMapping`'s `disableShift && !hostHandlesUIDMapping` branch (added in #667/#668) then sets `raw.idmap` explicitly and the workspace mount succeeds without `shift=true`. But nothing connects the two: the error message gives no hint that `disable_shift` is the fix, and the config field's own doc comment only mentions Colima/Lima:

```go
DisableShift bool   `toml:"disable_shift"` // Disable UID shifting (for Colima/Lima environments)
```

So a user hitting this on OrbStack has to independently discover a config flag whose own documentation says it's for a different VM host.

## Suggested fix

A few options, roughly increasing in effort:

1. **Cheapest — fix the doc/comment.** Stop asserting OrbStack's shift mounts are known-good, and mention OrbStack explicitly in `DisableShift`'s doc comment so it's discoverable from the config schema itself.
2. **Better error message.** When the device-mount step fails with "idmapping abilities are required but aren't supported on system", catch it and point the user at `disable_shift = true` directly, instead of surfacing the raw Incus error.
3. **Best — auto-fallback.** Mirror the existing `security.idmap.isolated` retry-without-isolation pattern (#534) at the device-mount level: on this specific error, retry the mount with `shift=false` + `raw.idmap` instead of leaving the container launch to fail outright.

Happy to send a PR for (1) or (2) if useful — wanted to flag the repro and the two root causes (stale assumption in `vmhost.go`, undiscoverable escape hatch in `DisableShift`'s doc comment) first since (3) touches the retry/launch pipeline and I'd want to talk through the right shape before writing it blind.

## Workaround for now

`[incus] disable_shift = true` in `~/.coi/config.toml` (or the profile/global config). Already needed on this same host for the unrelated-but-adjacent #667/#668 `raw.idmap` gap, so most OrbStack users hitting one will already have it set.

## Environment

- `code-on-incus` commit `b43d7fe` (2026-08-01, current `master`)
- OrbStack 2.2.1, guest kernel `7.0.11-orbstack-00360-gc9bc4d96ac70`
- Incus, both `dir` and `btrfs` storage pools tested
- `~/.coi/config.toml`: `code_uid = 501` (matches host UID), `disable_shift = true` (the workaround, already applied)

mensfeld:
Fixes #678 (thanks @technicalpickles for the detailed repro + root-cause). Implements all three suggestions per @mensfeld.

## Problem

`vmhost.go` detects OrbStack and *assumed* its virtiofs mounts support Incus's idmapped (`shift=true`) mounts. So unless `disable_shift` is set by hand, coi sends `shift=true` on the workspace device. On OrbStack kernels that can't do idmapped mounts, that fails at **container start**:

```
Error: Failed to setup device mount "workspace": idmapping abilities are required but aren't supported on system
```

…with no fallback — the existing `security.idmap.isolated` retry (#534) is a *different* code path and doesn't help.

## Fix (all 3 suggestions)

1. **Docs** — `vmhost.go` no longer claims OrbStack's shift mounts are known-good (it's kernel-dependent); `DisableShift`'s config doc now mentions OrbStack, not just Colima/Lima.
2. **Better error** — when the fallback can't recover, the raw Incus error is wrapped with a hint pointing at `[incus] disable_shift = true`.
3. **Auto-fallback (reactive)** — on that specific start failure, coi converts the container's `shift=true` disk devices to `shift=false` and sets `raw.idmap = both <hostUID> <codeUID>` (the exact mapping a manual `disable_shift` produces via `decideUIDMapping`), then retries. Reactive by design, so it works regardless of a given OrbStack kernel's capability instead of re-hardcoding a guess — try shift, drop to raw.idmap only if start actually fails.

Wired into **both** start paths: `StartWithIsolationFallback` (`coi shell` / persistent reuse) and `LaunchContainerWithPreStart`'s ephemeral path (`coi run`). Start calls now go through a stderr-capturing helper so the failure reason is matchable (plain `IncusExec` sends stderr to the terminal and returns a bare exit error).

## Tests & verification

- Unit tests for the detection (`isIdmapMountUnsupported` — matches the Incus message case-insensitively, wrapped or not; ignores unrelated failures) and the hint (`withDisableShiftHint` — preserves `errors.Is`, mentions `disable_shift`).
- `go build`/`vet`/`gofmt`/`golangci-lint` all clean; full `internal/container`, `internal/vmhost`, `internal/config`, `internal/session` suites pass.
- **CI limitation (called out honestly):** the full fallback flow requires a host whose kernel can't do idmapped mounts (OrbStack) — CI's Ubuntu+incus *does* support them, so the failure can't be reproduced there. The device-reconfigure + retry glue is thin over `incus config device set` / `config set` and is covered by logic + review; the decision surfaces (detection, hint, raw.idmap value = same as `decideUIDMapping`) are unit-tested.

CHANGELOG: `0.11.1 (Unreleased)` entry added.


mensfeld:
@technicalpickles can you test this out?

technicalpickles:
Tested this branch (`b4ee6ae`) on the host from #678 — OrbStack 2.2.1, guest kernel `7.0.11-orbstack-00360-gc9bc4d96ac70`, with `[incus] disable_shift` removed from my config so the fallback is what's actually under test.

## Sharper repro than the one in #678

The failure is **filesystem-specific, not host-specific**. Same host, same coi, same config — only the workspace path changes:

| workspace | filesystem | `shift=true` start |
|---|---|---|
| `/tmp/...` | tmpfs | works |
| `/home/...` | btrfs (guest's own disk) | works |
| `/Users/...` | virtiofs (macOS passthrough) | **fails** |

So this OrbStack kernel *can* do idmapped mounts — just not over the virtiofs share, which is exactly where coi workspaces live for anyone editing files on the Mac side. Worth recording because it's a good argument for this PR's reactive design over any static per-VM-host capability guess: a hardcoded "OrbStack can" *or* "OrbStack can't" would both be wrong on this machine.

## What works

- **`coi run` (ephemeral)** — fallback fires, prints the warning, container starts. `/workspace` shows `code:code` inside; files created in-container land as `501:501` on the host. Exit 0.
- **`coi shell` on a fresh persistent container** — same, and the resulting container has `raw.idmap: both 501 501`, all five disk devices flipped to `shift=false`, and `security.idmap.isolated: "true"` **preserved** (running the new branch before the isolation unset is the right ordering — nice).
- `TestIsIdmapMountUnsupported` / `TestWithDisableShiftHint` pass; `internal/session` suite passes (461s) on this host.

## Gap: the persistent-container **reuse** path still hard-fails

`internal/session/setup.go:285` (the "Starting existing container..." branch) calls `result.Manager.Start()` directly — no fallback, no hint. Reproduced with a plain user-facing flag against a stopped persistent container:

```
$ coi shell --slot 3 --background --debug
[setup] Starting existing container...
Error: Failed to setup device mount "protect-claude-settingsjson": idmapping abilities are required but aren't supported on system
Error: setup-container: failed to setup session: failed to start container: exit status 1
```

And it's not just "the fallback doesn't run on reuse". A few lines up, `applySessionSecurity(result.Manager, opts, reuseCWP, !opts.DisableShift, ...)` (setup.go:277) strips and re-adds the `protect-*` devices with **`shift=true` on every reuse**. So a container this PR's fallback already healed gets re-broken each session and then dies at start with nothing to catch it. Reachable via `--slot`, `--container`, `--resume`, and any `[container] persistent = true` setup.

## Implementation notes for the follow-up

Three call sites start a container without the new fallback:

1. **`internal/session/setup.go:285`** — persistent/`--container` reuse of a *stopped* container. This is the one that bites in practice.
2. **`internal/session/setup.go:558`** — the `else` of `if idmapIsolated` on the fresh-launch path; anyone with idmap isolation off gets no fallback on first launch either.
3. **`internal/cli/container.go:57`** — `coi container start <name>`. Minor, but same one-liner.

Smallest correct fix for (1) is a one-line swap to `container.StartWithIsolationFallback(result.ContainerName)` — which is exactly what `internal/cli/run.go:203` already does for `coi run --persistent`'s reuse, so it also makes the two reuse paths symmetric. It's safe on a non-isolated container: the isolation branch only runs *after* the idmap fallback has been tried, `config unset` on an unset key is a no-op, and `wasIsolated` comes back false so nothing gets re-set. The only wart is the "UID namespace isolation not available" warning showing up for unrelated permanent failures — already true of the fresh isolated path today.

For (2), that wart matters more (the container never had isolation), so it's worth splitting the new logic out of `StartWithIsolationFallback` into a standalone

```go
// StartWithIdmapFallback starts a container with only the #678 shift→raw.idmap
// recovery, for callers that never set security.idmap.isolated.
func StartWithIdmapFallback(containerName string) error
```

— start, poll `ContainerRunning`, and on `isIdmapMountUnsupported(firstErr)` do `fallbackShiftToRawIdmap` + retry, wrapping with `withDisableShiftHint` if the retry still fails. `StartWithIsolationFallback` then calls it first and keeps its isolation logic for the residual failure. ~25 lines, all of it moved rather than new.

Optional third piece: stop re-arming the breakage. In the reuse branch, `useShift` could be `!opts.DisableShift && !containerUsesRawIdmap(containerName)`, where the helper is a `incus config get <name> raw.idmap` non-empty check. Without it the reactive fallback still recovers, it just eats a failed start plus a retry on every single session. Cheap and unit-testable, but strictly a nice-to-have on top of the real fix.

Happy to send this as a follow-up PR if you'd rather keep this one as-is — just say the word so we don't both write it.

## Net

The two paths this PR wires up genuinely work here — this is the first time `coi run` and a fresh `coi shell` have started on this host without `disable_shift`. I'm keeping `disable_shift = true` in my config for now only because of the reuse gap above.

technicalpickles:
Follow-up on the detection question, since "why react to a failed start instead of just asking the kernel first?" is the obvious next thought and I wanted a real answer rather than a guess.

## The exact check exists, but needs root

Incus asks for a `shift=true` device by making an idmapped mount: `open_tree(OPEN_TREE_CLONE)` then `mount_setattr(MOUNT_ATTR_IDMAP)`. You can ask that question directly without starting anything. Results on the #678 host (OrbStack 2.2.1, kernel `7.0.11-orbstack`):

```
running as uid 501   (the uid coi actually runs as)
/Users   open_tree      -> operation not permitted
/        open_tree      -> operation not permitted
/tmp     open_tree      -> operation not permitted

running as uid 0
/Users   mount_setattr  -> invalid argument      <- virtiofs refuses idmapping
/        mount_setattr  -> OK                    (btrfs)
/tmp     mount_setattr  -> OK                    (tmpfs)
```

As root the probe is perfectly precise and matches the observed behaviour exactly. As the unprivileged user it can't even get past `open_tree`, so it can't distinguish "filesystem says no" from "you're not allowed to ask". `can_idmap_mount()` checks `ns_capable(fs_userns, CAP_SYS_ADMIN)` before it checks `FS_ALLOW_IDMAP`, so an unprivileged caller always gets `EPERM` first.

coi runs unprivileged against incusd, so it can't run this. And Incus doesn't expose a "can you idmap this path" API, so the only root-side probe available is "attach the device and try to start", which is what this PR already does reactively. `environment.kernel_features` doesn't help either: it's kernel-wide, and this kernel *does* support idmapped mounts, it just can't do them on that one filesystem.

<details>
<summary>probe source, if you want to run it on Lima/Colima</summary>

```go
func probe(path string, usernsFD int) {
	fd, err := unix.OpenTree(unix.AT_FDCWD, path,
		uint(unix.OPEN_TREE_CLONE|unix.OPEN_TREE_CLOEXEC))
	if err != nil {
		fmt.Printf("%-8s open_tree      -> %v\n", path, err)
		return
	}
	defer unix.Close(fd)

	attr := &unix.MountAttr{
		Attr_set:  unix.MOUNT_ATTR_IDMAP,
		Userns_fd: uint64(usernsFD),
	}
	if err := unix.MountSetattr(fd, "", unix.AT_EMPTY_PATH, attr); err != nil {
		fmt.Printf("%-8s mount_setattr  -> %v\n", path, err)
		return
	}
	fmt.Printf("%-8s mount_setattr  -> OK\n", path)
}
```

`usernsFD` is `/proc/<pid>/ns/user` of a child spawned with `CLONE_NEWUSER` and a valid uid map (Go's `SysProcAttr.UidMappings` does it in a few lines).
</details>

## The cheap proxy that does work unprivileged

```
/Users: fuseblk
/home:  btrfs
/tmp:   tmpfs
```

One `statfs` on the disk device's source path, no privileges, no wasted container start. FUSE-family filesystems only permit idmapped mounts when the server negotiated it at mount time, and that negotiation isn't visible from outside the mount, so "source is FUSE" is not decidable in the general case. But treating it as "skip shift, use raw.idmap" is conservative in the safe direction: the outcome is always correct, you just occasionally use raw.idmap where shift would have worked.

What makes this appealing beyond fixing #678: it isn't OrbStack-specific. Lima and Colima share the host filesystem the same way, which is exactly why `vmhost.go` carries a hardcoded VM-host list and why `disable_shift` exists. A statfs on the actual source path replaces a guess about the VM with a property of the path, and the OrbStack special case stops being special.

Layered with what's already in this PR:

1. `statfs` the source up front, skip `shift=true` for FUSE-family sources (cheap, no failed start, covers Lima/Colima/OrbStack uniformly)
2. keep this PR's reactive fallback underneath as the backstop for anything the heuristic misses
3. `withDisableShiftHint` stays as the final surface if both fail

That's design input for whoever picks up the follow-up, not a request to grow this PR. The reactive fallback here is correct as-is and is what makes the host usable today. Worth knowing the numbers exist, so nobody has to re-derive them: the exact check needs root, the good-enough check is one syscall.

technicalpickles:
Well, this is annoying timing. OrbStack shipped 2.2.2 on Aug 2, I upgraded, and re-ran everything from my first comment.

The error this PR matches on doesn't happen anymore on current OrbStack.

## What changed

2.2.2 turned on idmapped mounts for the Mac share. Probing `mount_setattr(MOUNT_ATTR_IDMAP)` directly, as root:

```
v2.2.1:  /Users -> EINVAL     v2.2.2:  /Users -> OK
```

So `shift=true` device mounts attach now, and the container starts. The mapping they produce is wrong, though.

## The new failure

Stock coi, `disable_shift` removed, workspace on the Mac share, OrbStack 2.2.2:

```
uid=501(code) gid=501(code)
-rw-r--r-- 1 65534 65534 10 Aug  4 23:25 f.txt
touch: cannot touch '/workspace/written-in-container': Permission denied
```

The launch reports success. Every write into `/workspace` then fails. Same coi, same config, workspace on btrfs instead: `501 501`, writes fine.

The cause is on OrbStack's end and I've written it up at orbstack/orbstack#2530. Short version: their share reports ownership based on the asking uid, so every process sees itself as the owner. An idmapped mount translates whatever uid the filesystem reports, so it maps an already-mapped value and lands on junk (the file reads as `1000000`, the container's host idmap base). Control test on the same container, `shift=false`:

```
asker uid    mac share      btrfs control
0            uid=0          uid=65534
1000         uid=1000       uid=65534
65534        uid=65534      uid=65534
```

## What that does to this PR

`isIdmapMountUnsupported` matches on `"idmapping abilities are required"`, and that message is gone on 2.2.2. So:

- `fallbackShiftToRawIdmap` never fires
- `withDisableShiftHint` never fires
- someone on current OrbStack without `disable_shift` gets a container that starts clean and a workspace they can't write to, with nothing anywhere telling them why

I still think this should land. It's correct for 2.2.1 and older, it's correct for any host that fails loudly, and the hint is an improvement on a raw Incus error. I just didn't want it going in under the impression that it closes out OrbStack, because as of two days ago it doesn't.

## Which promotes the statfs idea

In my last comment I pitched `statfs`-ing the disk device's source and skipping `shift=true` for FUSE-family sources, framed as an optimization to save a wasted container start. It's not an optimization anymore. It's the only thing left that works, because there's no error to react to. The call has to happen before start, off a property of the source path.

It also handles both eras with one rule. The loud 2.2.1 failure and the silent 2.2.2 one come from the same source filesystem, and neither happens if you don't ask for `shift` on it. Same for Lima and Colima, which is what the hardcoded list in `vmhost.go` is approximating today.

So same shape as before, with the reactive layer demoted:

1. `statfs` the source, skip `shift=true` for FUSE-family sources (primary)
2. this PR's fallback underneath as the backstop, and it's what covers older OrbStack
3. `withDisableShiftHint` as the last surface

The reuse-path gap from my first comment (`internal/session/setup.go:285` calling `Manager.Start()` with no fallback) is unrelated to any of this and still stands.

One more data point while I was in there: `[incus] disable_shift = true` is still the right config on OrbStack of any version. On 2.2.1 because shift fails loudly, on 2.2.2 because it fails quietly. I re-verified the `raw.idmap` path on 2.2.2 and it's fine, `501 501` in the container and writes land on the host owned correctly.

mensfeld:
@technicalpickles dude so what is the conclusion here? :sweat_smile:

Issue #686, opened 2026-08-06
coi health: non-thin LVM pools (lvm.use_thinpool=false) behave like dir but aren't flagged


mensfeld:
Follow-up from the #659 driver check (#684). Incus's LVM driver with `lvm.use_thinpool=false` is, per Incus's own docs, "close to the `dir` driver both in speed and storage usage": no CoW snapshots, so every launch pays a full image copy — the exact cost the dir warning exists for — yet the driver string is `lvm` and `coi health` reports the pool OK.

The same holds structurally for `lvmcluster` (cannot use thin pools). `cephfs`/`cephobject` don't host instances, so `dir` + non-thin LVM are the only real gaps.

Why it wasn't done in #684: `lvm.use_thinpool` is only visible via `incus storage show <pool>` (pool config), not `storage list`/`storage info`, so detection costs one extra incus invocation per LVM pool. The flag defaults to `true` and the docs discourage disabling it, so the affected population is small.

Suggested shape: only when a pool's driver is `lvm` (or `lvmcluster`), run `incus storage show` and warn when `lvm.use_thinpool: "false"` (lvmcluster: always). Zero extra cost for the common zfs/btrfs/dir cases.

Issue #659, opened 2026-07-27
coi run startup is ~24s, 73% of it one incus init re-unpacking the image (dir storage pool)


technicalpickles:
A no-op `coi run -- true` — launch a container, run `true`, tear it down — takes **~24s warm / ~31s cold** here. I instrumented the run pipeline to find out where that goes, and it is dominated by a single call: `incus init` re-unpacks the entire image into a new container directory on every launch, because the storage pool uses the `dir` driver.

## Numbers

`coi run -- true`, 3 consecutive runs, nothing else touching Incus (median):

| bucket | median | share |
|---|---:|---:|
| `incus init <image> <name> --ephemeral` | 17.8s | **73%** |
| `incus start <name>` | 3.1s | 13% |
| `incus delete <name> --force` (teardown) | 1.3s | 5% |
| 41 other `incus` calls (config set, device add, exec, list …) | 1.3s | 5% |
| coi's own Go code | 0.9s | 4% |
| **total** | **24.3s** | |

Per pipeline phase (one representative run, total 24.2s):

```
phase      validate-env           0.04s
phase      launch-container      21.33s     <- 88%
phase      configure-container    0.38s
phase      apply-network          0.35s
phase      start-monitoring       0.00s
phase      run-command            0.02s
teardown   apply-network          0.76s
teardown   launch-container       1.36s
```

So **96% of wall time is `incus` subprocesses**, and 73% of it is one of them. Dropping the page cache first (`sysctl -w vm.drop_caches=3`) moves the total to 31.1s with `incus init` at 22.1s — same shape, more I/O.

## Why `incus init` costs 18s

The pool is the stock `dir` driver:

```
$ incus storage list --format=csv
default,dir,,5,CREATED
```

With `dir`, Incus keeps no unpacked image volume to clone from — the pool's image cache directory stays empty across launches:

```
$ sudo ls -la /var/lib/incus/storage-pools/default/images/
total 0        # empty, before and after a launch
```

…so every `incus init` unpacks the full image tarball again. Incus says so itself during the call: `Retrieving image: Unpacking image: 45% (51.40MB/s)`.

It scales with unpacked image size, as you'd expect from a copy — bare `incus init`, no coi involved:

| image | compressed | unpacked | files | `incus init` |
|---|---:|---:|---:|---:|
| `coi-ubuntu-24.04` | 253 MiB | 1.2 GB | 47k | 3.2s / 5.7s |
[...]

PR #687, opened 2026-08-06
fix(health): source pool driver from storage list JSON + harden the e2e test (#684 follow-up)

mensfeld:
Applies the post-merge code-review findings on the #659/#684 storage-pool driver check.

## Product changes

- **Driver from structured JSON.** The driver now comes primarily from `incus storage list --format=json` via the existing `container.ListStoragePools()` helper (already used by `coi clean`). The text scrape of `incus storage info` had no failure signal — a future Incus output reshape would have silently disabled the very warning the check exists for — and remains only as a fallback.
- **Consistent per-pool schema.** The error-path details entry now carries the `driver` key too (it's known independently of the usage query), and a `dir` pool still surfaces its warning when usage is unparseable.
- **Consistent message.** Usage line first, dir warning after it, both under the same `pool (driver)` label:

```
[WARN] Incus storage pools: default (dir): 77.3 GiB free of 80.2 GiB (4% used); default (dir): 'dir' storage driver re-unpacks the image on every launch — ...
```

## Test changes

- **e2e test hardened for CI reality**: tolerates a per-pool `failed` entry (unparseable usage query) and warning/failed usage thresholds (a dir pool mirrors the root filesystem, which can be nearly full on runners) while still asserting the driver and warning in every state; `entry.get()` instead of bare subscripts.
- **Single `coi health` invocation**: the text renderer prints `check.Message` verbatim, so the message assertions on the JSON run cover the text output — the second full battery run is dropped.
- **Helper dedup**: `is_incus_permission_error` / `create_storage_pool` / `delete_storage_pool` hoisted into `tests/support/helpers.py`, replacing 3–4 verbatim copies across `tests/health/`.
- New Go unit cases: driver from the structured list, text fallback when the list is unavailable, list-wins-over-fallback, and the error-path-keeps-driver + warning behavior. Warning-follows-usage ordering asserted.

## Verification

- `go test ./internal/health/`, `go vet`, `golangci-lint`: clean; `ruff check` + `ruff format --check`: clean.
- All four touched Python tests pass against a live Incus with a real `dir` pool; text output verified end-to-end (sample above).

Deferred finding: non-thin LVM pools (`lvm.use_thinpool=false`) behave like `dir` but aren't flagged — tracked in #686 (needs an extra `incus storage show` per LVM pool; the flag defaults to on).


mensfeld:
Second-pass review applied (8 findings from a recall-focused review of this branch itself):

- **Error path no longer self-contradicts**: an enumerated pool with a failed usage query reports `pool (driver): usage unavailable: <err>` instead of `pool: missing` next to its own driver warning. Truly unresolvable pools still say missing.
- **`evaluatePool()` extraction**: both loop paths flow through one per-pool function, so driver-based diagnoses (dir today, non-thin LVM #686 tomorrow) can't drift between the error and success paths.
- **Tests genuinely pin the usage line again** — the shared `pool (driver)` label had made the warning line alone satisfy both suites' asserts (the Go ordering check was vacuous when the usage line was absent). Go now requires both lines present and ordered; the e2e test binds the warning to its own pool (a host whose default pool is also dir can't mask a regression) and pins the labeled usage line by regex.
- **No more threshold duplication in the e2e test**: the asserted property is "a dir pool is never ok"; warning-vs-failed stays in Go unit tests.
- **Helper dedup completed suite-wide**: 5 `tests/clean/` files + `tests/build/build_with_storage_pool.py` migrated to the shared helpers with the documented contract (skip only on permission error, fail loud otherwise) and random-suffix pool names — a leaked pool can no longer permanently skip a test.

Verified locally: full Go suite + lint clean; all 9 touched Python tests pass against a live Incus with real dir pools.

Issue #667, opened 2026-07-28
raw.idmap still skipped when code_uid is manually set to match the host UID (gap in #530's fix)


technicalpickles:
## What's going on

`decideUIDMapping` (`internal/session/setup_helpers.go:69`) skips `raw.idmap` entirely whenever `hostUID == codeUID`:

```go
func decideUIDMapping(hostUID, codeUID int, disableShift, hostHandlesUIDMapping bool) (useShift bool, idmap string) {
	if !disableShift && hostHandlesUIDMapping {
		disableShift = true
	}
	if hostUID != codeUID {
		return false, fmt.Sprintf("both %d %d", hostUID, codeUID)
	}
	return !disableShift, ""
}
```

That's the right call when shift is on, or on a real Colima/Lima guest where the VM's own user already sits at the host UID. But when `disable_shift` is manually set (my case: OrbStack, where shift is unsupported for an unrelated reason, tracked/worked around per #553) and `code_uid` is set to match the host UID as the documented workaround from #530, this branch returns no `raw.idmap` *and* no shift. The container falls back to its default unprivileged subuid allocation (e.g. host `1065536-1131071` → container `0-65535`), which doesn't include the real host UID at all. From inside the container, every workspace file shows up owned by `nobody:nogroup` (65534), and `/workspace` is completely unwritable.

## How I found it

Repro on OrbStack (Incus container, `code_uid = 501` + `disable_shift = true` in `~/.coi/config.toml`, host user is uid 501):

```
$ stat -c "%U %u %G %g" ~/projects/some-workspace
technicalpickles 501 technicalpickles 501

$ incus exec <container> -- stat -c "%U %u %G %g" /workspace
nobody 65534 nogroup 65534

$ incus exec <container> -- touch /workspace/.test
touch: cannot touch '/workspace/.test': Permission denied
```

`incus config show <container>` confirms no `raw.idmap` is set anywhere on the container or the `workspace` device, matching what `decideUIDMapping` returns when `hostUID == codeUID`.

Manually forcing it fixes it completely:

```
$ incus stop <container>
$ incus config set <container> raw.idmap "both 501 501"
$ incus start <container>

$ incus exec <container> -- stat -c "%U %u %G %g" /workspace
code 501 code 501

$ incus exec <container> -- touch /workspace/.test && echo WRITABLE
WRITABLE
```

Not durable, coi recreates the workspace device without `raw.idmap` on the next `coi run`/`coi shell` (same caveat as #553).

[...]

Issue #685, opened 2026-08-06
`coi shell` on a stopped persistent container skips #679's idmap fallback, and every reuse re-arms shift=true


technicalpickles:
#679 taught coi to recover when a `shift=true` device mount fails at container start, and wired it into `coi run` and fresh `coi shell`. The persistent-container **reuse** path never got it. Worse, that path actively re-breaks containers the fallback already healed.

Found while verifying #679 ([details in that thread](https://github.com/mensfeld/code-on-incus/pull/679#issuecomment-5185170625)). This is independent of the OrbStack-version stuff in #683: it bites on any host where shift mounts don't work, whatever the reason.

## Repro

Stopped persistent container, no `disable_shift`, on a host whose workspace filesystem can't do idmapped mounts (OrbStack 2.2.1 in my case):

```
$ coi shell --slot 3 --background --debug
[setup] Starting existing container...
Error: Failed to setup device mount "protect-claude-settingsjson": idmapping abilities are required but aren't supported on system
Error: setup-container: failed to setup session: failed to start container: exit status 1
```

Reachable via `--slot`, `--container`, `--resume`, and any `[container] persistent = true` config.

## Two things going wrong

**1. No fallback on reuse.** `internal/session/setup.go:285` calls `result.Manager.Start()` directly. The #679 recovery is sitting right there in `internal/container` and never runs, so you get the raw Incus error with no hint either.

**2. Every reuse re-arms the breakage.** A few lines up, `internal/session/setup.go:277` passes `!opts.DisableShift` to `applySessionSecurity`, which strips and re-adds the `protect-*` devices with `shift=true`. So a container that #679's fallback already converted to `raw.idmap` gets handed the broken config again each session, and then dies at start with nothing to catch it.

Two more call sites start without the fallback. Lower stakes, same one-liner:

- `internal/session/setup.go:558`, the `else` of `if idmapIsolated` on the fresh-launch path. Anyone running with idmap isolation off gets no fallback on first launch either.
- `internal/cli/container.go:56`, `coi container start <name>`.

(Line numbers as of 6d3fcf7.)

## Fix

[...]

PR #689, opened 2026-08-06
fix(uidmap): skip shift for FUSE-backed mount sources before start (#683)

### LINKED ISSUE #683: OrbStack 2.2.2 breaks the workspace silently under shift=true — needs proactive statfs-based skip (#679 fallback can't fire)
Follow-up to #678 / #679. On **OrbStack 2.2.2+** the workspace breaks *silently* when `shift=true` is used on the Mac share — the reactive fallback merged in #679 cannot fire because there is no start error to react to.

All findings below are from @technicalpickles' verification runs on the #679 PR (OrbStack 2.2.1 → 2.2.2, guest kernel `7.0.11-orbstack`).

## What changed in OrbStack 2.2.2 (shipped 2026-08-02)

2.2.2 enabled idmapped mounts on the macOS virtiofs share. Probing `mount_setattr(MOUNT_ATTR_IDMAP)` as root:

```
v2.2.1:  /Users -> EINVAL     v2.2.2:  /Users -> OK
```

So on ≤2.2.1 a `shift=true` device mount **fails loudly** at container start ("idmapping abilities are required but aren't supported on system") and #679's fallback catches it, converts to `raw.idmap`, and retries. On 2.2.2 the same mount **attaches successfully** — and produces garbage ownership.

## The silent failure

Stock coi, no `disable_shift`, workspace on the Mac share, OrbStack 2.2.2:

```
uid=501(code) gid=501(code)
-rw-r--r-- 1 65534 65534 10 Aug  4 23:25 f.txt
touch: cannot touch '/workspace/written-in-container': Permission denied
```

The launch reports success; every write into `/workspace` then fails, with nothing anywhere telling the user why. Same coi + config with the workspace on btrfs: `501:501`, writes fine.

Root cause is on OrbStack's side (reported upstream as orbstack/orbstack#2530): their share reports ownership relative to the asking uid, so every process sees itself as the owner. An idmapped mount then translates an already-translated value and lands on junk (files read as `1000000`, the container's host idmap base).

Consequences for the #679 machinery on 2.2.2:

- `isIdmapMountUnsupported` never matches (the message no longer occurs)
- `fallbackShiftToRawIdmap` never fires
- `withDisableShiftHint` never fires

## Fix direction: proactive statfs check (no longer an optimization)

Because there is no error to react to, the decision has to happen **before** start, off a property of the disk device's *source path*. Proposed in the #679 discussion:

1. **Primary:** `statfs` the source of each disk device; for FUSE-family filesystems (OrbStack's share reports `fuseblk`), skip `shift=true` and use the `raw.idmap` path directly. Unprivileged, one syscall, conservative in the safe direction (worst case: raw.idmap used where shift would have worked). Not OrbStack-specific — Lima and Colima share the host filesystem the same way, so this replaces the hardcoded VM-host guess in `vmhost.go` with a property of the actual path.
2. **Backstop:** keep #679's reactive fallback (covers OrbStack ≤2.2.1 and any other host that fails loudly).
3. **Last surface:** `withDisableShiftHint` if both fail.

Notes from the probe work (so nobody re-derives it): the *exact* kernel check (`open_tree` + `mount_setattr(MOUNT_ATTR_IDMAP)`) requires root — unprivileged callers get `EPERM` at `open_tree` before the filesystem is ever consulted — and Incus exposes no "can you idmap this path" API; `environment.kernel_features` is kernel-wide and useless here (the 2.2.1 kernel *does* support idmapped mounts, just not on virtiofs).

## Related but separate

The persistent-**reuse** start path (`internal/session/setup.go:285`) still calls `Manager.Start()` with no fallback, and `applySessionSecurity` re-adds `protect-*` devices with `shift=true` on every reuse — re-breaking containers the #679 fallback healed. @technicalpickles mapped the fix and offered a follow-up PR in the #679 thread; that gap deserves its own PR/issue and is not blocked on this one.

## Workaround (all OrbStack versions)

```toml
[incus]
disable_shift = true
```

Right on 2.2.1 because shift fails loudly; right on 2.2.2 because it fails silently. The `raw.idmap` path was re-verified on 2.2.2: `501:501` in the container, host writes land correctly owned.

technicalpickles:
Fixes #683.

## Why the reactive fallback isn't enough anymore

#679 recovers by waiting for the container start to fail on an idmapped mount, converting the shift devices to `raw.idmap`, and retrying. That works, and it's still the right backstop. But OrbStack 2.2.2 (shipped Aug 2) enabled idmapped mounts on the macOS share, which removed the failure it keys on without making the mount actually correct:

```
uid=501(code) gid=501(code)
-rw-r--r-- 1 65534 65534 10 Aug  4 23:25 f.txt
touch: cannot touch '/workspace/written-in-container': Permission denied
```

The share reports ownership relative to the asking uid, so an idmapped mount translates an already-translated value and lands on junk (files read as `1000000`, the container's host idmap base). Root cause is on OrbStack's side, reported upstream as orbstack/orbstack#2530.

The launch reports success. `isIdmapMountUnsupported` never matches, `fallbackShiftToRawIdmap` never fires, `withDisableShiftHint` never fires, and the user gets a container that starts clean and a workspace they can't write to with nothing anywhere explaining it. There's no error left to react to, so the decision has to move before start.

## The fix

`statfs` the disk device's source path and skip `shift=true` for FUSE-family filesystems. One syscall, no privileges, no wasted container start.

The exact check exists but needs root: Incus makes an idmapped mount with `open_tree(OPEN_TREE_CLONE)` then `mount_setattr(MOUNT_ATTR_IDMAP)`, and `can_idmap_mount()` tests `ns_capable(fs_userns, CAP_SYS_ADMIN)` before it ever consults `FS_ALLOW_IDMAP`. So an unprivileged caller gets `EPERM` at `open_tree` and can't distinguish "this filesystem refuses idmapping" from "you aren't allowed to ask". coi runs unprivileged against incusd. Incus exposes no "can you idmap this path" API, and `environment.kernel_features` is kernel-wide and useless here, since these kernels do support idmapped mounts, just not on this filesystem.

"Source is FUSE" doesn't decide the question either, strictly speaking: a FUSE filesystem permits idmapped mounts when its server negotiated `FUSE_ALLOW_IDMAP` at mount time, and that negotiation isn't visible from outside the mount. But it's wrong only in the harmless direction. `raw.idmap` produces a correct mapping either way, so a false positive costs an optimization while a false negative costs a container that won't start (#678) or one whose workspace is silently unwritable (#683). A path that can't be stat'd returns false, leaving the caller on its existing behaviour.

The nice part is it isn't OrbStack-specific. Lima and Colima share the host filesystem the same way, which is what the hardcoded VM-host list in `vmhost.go` has been approximating. A property of the actual source path beats a guess about the VM around it, and it handles both eras of this bug with one rule: the loud ≤2.2.1 failure and the silent 2.2.2 one come from the same source filesystem, and neither happens if you don't ask for shift on it.

Layering, unchanged from what #683 proposed:

1. `statfs` the source, skip `shift=true` for FUSE-family sources (this PR)
2. #679's reactive fallback underneath, still covering OrbStack ≤2.2.1 and anything the check misses
3. `withDisableShiftHint` as the last surface

## Shape of the change

`vmhost.SourceBlocksIdmappedMounts(path)` is the new check, with the filesystem table split into a pure `magicBlocksIdmappedMounts` so it's testable without needing one of these mounted.

It's a whole-container decision, not per-device: `shift` is per-device but `raw.idmap` is per-container and Incus won't combine them, so one FUSE-backed source among the disk devices means every device has to go the raw.idmap route. `vmhost.FirstBlockingSource(paths...)` returns the offending path (so the log can name it, which matters when several mounts are configured), and `session.MountSources(workspace, mountConfig)` builds the list: workspace first, then any configured `[[mount]]` host paths. Protected paths and secret masks aren't in the list because they live inside the workspace, so its filesystem already speaks for them.

That last part is why the list exists rather than just the workspace path. A workspace on the guest's own disk with a `[[mount]]` from the Mac share hits exactly the same silent breakage, and checking only the workspace would miss it.

The result feeds `decideUIDMapping` as a new input that folds into `disableShift` first, so the rest of the existing matrix applies unchanged. That matters for one case in particular: it reaches the #667 branch, so a host/code UID match on a FUSE source still gets `raw.idmap` rather than nothing, because the container's default subuid range doesn't cover hostUID just because the nominal code UID matches it.

`ConfigureUIDMapping` takes the workspace path now. New `ResolveUseShift(workspacePath, disableShift)` returns just the flag from the same decision, for the callers that need it before the container exists (the run pipeline's pre-start default) or on a reused container whose `raw.idmap` was applied at creation. That replaces the two remaining raw `!cfg.Incus.DisableShift` reads in `phases_run.go`, which were the last places the flag was decided without going through the shared function.

`disable_shift` stays as the manual override, but its config doc now says it's rarely needed.

One place I deliberately did **not** change, having tried it and been wrong: `coi health`'s secret-mask probe still hardcodes `useShift = true`. Its comment claimed it "matches COI's default workspace mount", which this PR makes untrue, so my first pass routed it through `ResolveUseShift`. That breaks the probe outright:

```
Error: Failed to start device "mask-e9cbb0224c4a3d23": Failed to add mount for device inside container
```

`shift=false` is only half of the non-shift path. The other half is `raw.idmap`, which takes effect at container START, and this probe has already started its container by the time it mounts. Handing it the real decision without the mapping gives a broken mount rather than a working one. Its workspace is a throwaway `MkdirTemp` on local storage anyway, so shift is the right answer for it regardless of what the user's real workspace would get. The comment now says that instead. Making the probe genuinely mirror a session would mean applying the mapping pre-start and is its own piece of work.

## Tests & verification

- `TestMagicBlocksIdmappedMounts` pins the filesystem table (fuse/fuseblk in; btrfs, tmpfs, ext4, xfs, overlay, zfs, 9p out).
- `TestSourceBlocksIdmappedMountsUnreadablePath` pins the fail-open behaviour, and `TestSourceBlocksIdmappedMountsRealDir` exercises the actual syscall against `t.TempDir()`, which is non-FUSE on any host CI might run on.
- `TestFirstBlockingSource` covers the list handling, `TestMountSources` covers how the list is built.
- `TestDecideUIDMapping` gains four cases for the new input, including the #667 interaction and the Colima/Lima case where the guest maps UIDs itself so no `raw.idmap` is needed.
- `go build` / `go vet` / `gofmt` clean.

On the full `-short` run, `internal/session`, `internal/cli`, `internal/vmhost`, `internal/config` and the rest pass. Three packages fail on my host and I chased each one down rather than waving at it:

- `internal/container`: `TestDirExistsAndFileExists_PathWithSpaces` and `TestSetTmpfsSize`, both at `launch container: exit status 1`. I ran the same two on master; they fail identically. Pre-existing here.
- `internal/health`: `TestCheckContainerConnectivity_WithImage` and `TestCheckNetworkRestriction_WithImage`, both failing their "no leftover containers" assertion and naming `coi-health-check-1785188577813546442` and `coi-restriction-check-1785189171438986455`. Both of those containers already existed on this host before I started, so they're leaked from an earlier run.
- `internal/network`: firewall assertions (`ping to gateway succeeded while boot block was in place`, a wildcard-in-allowed_domains case). `go list -deps ./internal/network` shows no dependency on `session`, `vmhost` or `health`, so nothing in this diff can reach it.

`internal/session` failed once on a whole-repo parallel run with `Failed to retrieve PID of executing child process` and passed on a serial re-run, which is daemon contention on a host where every container launch re-unpacks the image (#659).

**Verified against the real filesystem from the issue.** This host is OrbStack, guest kernel `7.0.14-orbstack`, with `code_uid = 501` matching the host uid so the UID-mismatch rule doesn't mask the result:

```
host uid=501, CodeUID=501, vmhost="orbstack"
workspace on btrfs, no extra mounts                  useShift=true
workspace on the Mac share                           useShift=false
workspace on tmpfs                                   useShift=true
workspace on btrfs + extra mount from the Mac share  useShift=false
workspace on btrfs + extra mount on tmpfs            useShift=true
```

The first three reproduce the table from my #679 comment exactly: the Mac share drops to `raw.idmap`, the guest's own btrfs and tmpfs keep `shift=true`. Those are the same three filesystems whose empirical behaviour motivated the issue, and the check now agrees with what each one actually does. The last two are the multi-source case.

Worth noting `/proc/mounts` calls `/Users` virtiofs while `statfs` reports `fuseblk`, which is why this keys on the statfs type rather than the mount line.

**One thing I could not test:** a FUSE filesystem that *does* negotiate `FUSE_ALLOW_IDMAP`, which is the false-positive case. I don't have one. The behaviour there is "uses raw.idmap where shift would also have worked", which is correct but not optimal, and it's the tradeoff the design deliberately accepts.

## Interaction with #688

They're independent and can merge in either order, with one wrinkle each way. #688 adds a `ConfigureUIDMapping` call to the persistent-reuse path; this PR changes that function's signature. Whichever lands second needs a one-line rebase, and I'm happy to do it.

Worth having both: this PR stops shift being asked for on a FUSE workspace, #688 stops the reuse path re-arming `shift=true` on every session and gives it the start fallback. Until #688 lands, the shell reuse path still decides its flag from `!opts.DisableShift` and won't see this check.


mensfeld:
Pushed a merge of master to resolve the conflicts (maintainer edit). Three things beyond the mechanical resolution:

1. **#688 interplay**: master's #685/#688 fix gave the persistent-reuse path its own `ConfigureUIDMapping` call — with this PR's old signature. The merge wires it to `ConfigureUIDMapping(containerName, MountSources(opts.WorkspacePath, opts.MountConfig), …)`, matching the fresh-launch call, so a **reused** container on a FUSE-backed workspace now gets the same pre-start raw.idmap decision as a fresh one, with #688's `reuseShiftDecision` folding in an existing raw.idmap on top. The two changes compose exactly as both intended.
2. **Latent lint/darwin conflict**: `golangci-lint`'s `unconvert` flags `int64(st.Type)` in `SourceBlocksIdmappedMounts` (on linux `Statfs_t.Type` is already `int64`) — but the conversion is load-bearing on darwin (`Type` is `uint32` there), which the release matrix builds. Kept the conversion with a `//nolint:unconvert` + comment; verified `GOOS=darwin` builds and lint is clean.
3. `uid_mapping_test.go` / CHANGELOG conflicts were purely additive — kept both sides (`TestMountSources` + `TestReuseShiftDecision`, both entries).

Full suite + lint green locally on the merged tree. Nice work on this one — the statfs approach and the fail-open reasoning are exactly right.

mensfeld:
Second maintainer pass (recall-focused review of the branch, run twice — findings applied in the two commits above):

**Round 1** closed coverage holes in the check itself: persistent-**reuse** now heals pre-upgrade containers on both pipelines (`session.ResolveReuseUIDMapping` applies the fresh-launch decision and proactively converts creation-time `shift=true` devices via `container.ConvertShiftedDiskDevices` — on OrbStack ≥2.2.2 the start failure the reactive fallback keys on never happens); the **git-worktree common dir** now votes (layout resolved before the decision, `WorktreeSources` feeding `MountSources`); a **not-yet-existing** mount source is judged by its nearest existing ancestor (that's where the later MkdirAll lands — verified against a real bindfs/FUSE mount); **9p** joins the table (a Lima `mountType: 9p` guest has none of `detect()`'s markers, and where real Colima/Lima is detected its veto still wins); run-reuse gains the raw.idmap-aware fold (#685 parity).

**Round 2** reviewed round 1 and corrected it: a trust pre-gate I'd added to the reuse vote was **reverted** — it recomputed a mounts-only fingerprint the combined-per-source trust store can never match, and on reuse the attached devices persist regardless of current trust anyway, so the ungated config is the truthful input (the vote can only flip toward raw.idmap, whose value is UID-derived — nothing injectable). Also: device conversion now runs only on the *transition* to raw.idmap (no more 1+2N incus calls on every reuse), `ConvertShiftedDiskDevices` reports failures instead of claiming success, strip-before-decide ordering matches across pipelines, worktree-resolution failures warn instead of silently dropping the vote, the dead pre-hook `ResolveUseShift` store is gone, and the FUSE log line is gated by re-asking `decideUIDMapping` itself rather than a hand-copied condition.

Deferred: `coi container start` still lacks the proactive decision (#691 — needs sources derived from the container's actual devices); unifying the two near-clone reuse blocks; narrowing the LimaLike virtiofs-substring detection.

Full suite + lint green locally at every step.

PR #692, opened 2026-08-10
[CI canary] pure master vs new incus daily build — do not merge

mensfeld:
Empty commit on top of master. PR #689's runs hit test_threat_deduplication failures twice in a row (~167s each) on the zabbly incus 7.3 Aug-7 daily build; the last green run of that test used the Aug-3 build. This canary determines whether the failure is master-level (environment) or PR-specific. Will be closed after the run.


mensfeld:
Canary served its purpose: full run green on the new zabbly incus build (202608072121), confirming master's baseline — the monitoring-response failures on #689 were the starvation-sensitive test_threat_deduplication flaking under runner load (passed on the third sample), and the misc-config limits cluster cleared on the first retry.

PR #693, opened 2026-08-10
feat(session): [container] session_name — named sessions that survive workspace moves

mensfeld:
Implements name-keyed session identity: `[container] session_name` (config/profile only — deliberately no CLI flag) decouples a session from its workspace location, so the same persistent session continues (`--resume`/`--continue`) from any checkout path.

## Design

- **One identity function**: `session.Identity(workspacePath, sessionName)` returns `session-name:<name>` when set (namespaced — a name can never collide with a real path), else the absolute path; `IdentityHash` feeds container names, slot allocation, port allocation, `ListWorkspaceSessions`, and the latest-session resume matcher. Saved sessions embed the hash in their recorded container name, so resume matching across workspaces required no metadata change.
- **Reuse remount**: a reused container's devices persist from creation, so launching a named session from a new location leaves the workspace device pointing at the old source. New `Manager.GetWorkspaceSource()` detects the mismatch and `session.RemountMovedWorkspace` replaces the workspace + `git-worktree-common` devices with the current source and shift decision, re-deriving the container-side path under the same preserve-path/worktree rules as a fresh launch — **before** the security overlays are re-applied, on both the shell and run pipelines. `[[mount]]`s are location-independent and keep their persisted sources.
- **Trusted scope only**, following the `[defaults] profile` precedent: the session name selects which persistent container and saved state (conversation history, restored credentials) a launch attaches to, so project-scoped (`./.coi`) values — top-level or in project profiles — are stripped with a warning.

## Testing

- Unit: identity keying (location-independence, path/name namespacing), remount device choreography via a narrow `WorkspaceRemounter` fake (no-op on same source, fail-open on unknown, replace order, preserve-path/worktree/system-dir rules, abort-before-mount on failure), config parse/merge/trust-strip/profile-apply.
- E2E (`tests/persist/named_session_across_workspaces.py`): one named persistent session continued across two different workspace dirs — same container (marker in container home survives), `/workspace` shows the new checkout after remount, old checkout gone.
- Verified live up to container start on incus 7.3 (the name-derived container `coi-<namehash>-1` and config threading confirmed with a real `coi run`; this sandbox can't start nested containers, so the runtime remount rides CI). Full Go suite, vet, golangci-lint, ruff clean.

The manual live check earned its keep: it caught that my first test round ran against a stale binary — and confirmed the final wiring produces the exact expected name-derived hash.


mensfeld:
Self-review round (4 finder angles, 24 candidates) applied as a hardening commit:

**Security**: profile `inherits` could smuggle a trusted parent's `session_name` into a project-scoped profile (the per-profile strip ran before inheritance resolution) — re-stripped on the resolved result, with a test.

**Correctness**: attaching to a *running* container still mounting a different workspace is now refused (can't remount live; silently working against the other checkout was the worst failure mode); the resume lookup + attach/monitor/snapshot now see a `[defaults]`-profile-carried name (they run before the profile fallback); pre-adoption path-keyed saved sessions still match `--resume`; busy-name auto-allocation warns loudly that it FORKS the session; symlink-aware move detection (no per-launch device churn via aliased paths); device source via `config device get` (the YAML re-parse could return quoted paths and treat every reuse as moved); undeterminable source logs instead of silently skipping; `session_name` validated at load (lookalike names with trailing/invisible chars would hash to different identities).

**Documented, deliberately not solved here** (in CHANGELOG): concurrent same-name launches can race the remount; two profiles sharing a name share one container regardless of image; `[[mount]]` sources under a moved workspace keep old paths; `preserve_workspace_path` changes the in-container path across checkouts so cwd-keyed tool history doesn't follow (the feature's history-continuity promise fully holds for the default `/workspace` mount). Also deferred: consolidating the container-side-path rule (now three call sites) into one helper — worth a follow-up, kept out of this PR to avoid churning the fresh-launch path late.

Verified: full Go suite + lint + ruff green; live check on incus 7.3 confirmed the name-derived container and config threading (nested sandbox can't start containers, so the remount itself rides CI).

PR #700, opened 2026-08-13
feat(tool): add OpenAI Codex CLI as a supported tool (#698)

### LINKED ISSUE #698: Use coi with OpenAI Codex
Hi @mensfeld 

Thanks for your great work on coi. This project is really helpful to me.

Until now, I used coi only with Claude and this worked very well. Now I got the opportunity to also test OpenAI Codex. I searched the repository and could only find a single mention of Codex, so I wonder what would be necessary to use Codex with coi and what the state of the integration is.

Since I have my own build.sh to create my environment, which is based on https://github.com/mensfeld/code-on-incus/blob/master/internal/image/build.sh, it was pretty straight forward to just install Codex side-by-side with Claude.

But now I do have some questions:

* I understand, that coi also handles some of the config files used by Claude. How would this work with Codex?
* How can I shell into Codex directly (without going through the bash shell)
* How to make configuration options like `tool.permission_mode = "bypass"` work with Codex

mensfeld:
Closes #698.

Adds OpenAI Codex CLI as a fourth supported tool, at parity with the claude/opencode/pi integrations: `[tool] name = "codex"` launches straight into codex, `permission_mode` maps to the right codex flags, host config is seeded, and the sandbox context lands in codex's native `AGENTS.md`.

## How it works

| coi concept | codex mapping |
|---|---|
| `permission_mode = "bypass"` (default) | `codex --dangerously-bypass-approvals-and-sandbox` — the Incus container IS the sandbox (codex's own Landlock sandbox may not work nested anyway), and the flag also skips the first-run folder-trust prompt |
| `permission_mode = "interactive"` | `codex -s workspace-write -a on-request` (codex keeps its approval prompts) |
| Credential/config seeding | host `~/.codex/{auth.json, config.toml, AGENTS.md}` via the same catalog bundle mechanism claude uses |
| `[tool.codex] model` / `reasoning_effort` | `-m <model>` / `-c model_reasoning_effort=<v>` |
| Auto-context | COI-managed block in container-global `~/.codex/AGENTS.md` (never the workspace `AGENTS.md`, which lives on the host bind-mount) |
| Resume | newest `sessions/**/rollout-*.jsonl` UUID → `codex resume <uuid>`, falling back to `codex resume --last` (layout-drift safety net); sessions live inside `~/.codex`, so the existing config-dir save/restore persists them for free |

## Design decisions

- **No settings-file injection for codex.** Codex config is TOML while coi's settings merge (`mergeJSONSettings`) is JSON-only, so the codex bundle has no `sandbox_settings_file`: the host `config.toml` is copied **verbatim**, and everything coi controls travels as launch flags — all centralized in `CodexTool.BuildCommand`, one fix site if upstream flags drift.
- **Opt-in at image build.** `install_codex` is added to build.sh (native installer with `CODEX_NON_INTERACTIVE=1`, IPv4 preference, 3× retry, binary verify, `/usr/local/bin` symlink), but the default `COI_AGENTS` set stays `claude opencode pi` — opt in with `[container.build] agents = ["claude", "codex"]`. A new `tool.DefaultBuildAgents()` (kept in lockstep with build.sh's `${COI_AGENTS:-...}` default by the dispatch test) extends the #454 footgun warning so `name = "codex"` + an unset agents list warns at build time instead of failing with exit 127 at launch.
- **Security: workspace `.codex/config.toml` joins `protected_paths`** (read-only mount + empty placeholder materialization when absent), mirroring the `.claude/settings.json` planting guard from #504 — project-scoped codex config can name commands a later host codex session would auto-execute (`notify`, MCP launchers).
- **Latent bug fixed along the way:** `getConfiguredTool` fed `[tool.claude]` model/effort into *any* tool implementing `ToolWithModel`/`ToolWithEffortLevel`; it's now tool-aware, with leak-regression tests in both directions.

## Auth

coi seeds the host's `~/.codex/auth.json`, so log in on the host first. If the host uses the OS keyring (no `auth.json`) or has never logged in, authenticate inside the container with `codex login --device-auth` (needs org enablement) or `codex login --with-api-key` — the plain browser flow can't work inside the container because its OAuth localhost callback is unreachable from the host browser (documented in the README). Host→container callback forwarding could be a follow-up if there's demand.

## Tests

- **Unit** — `internal/tool/codex_test.go` (full `BuildCommand` matrix, rollout-UUID discovery fixtures incl. malformed-name/tie-break/missing-dir cases, capability interface assertions), catalog bundle shape, registry, `[tool.codex]` merge precedence, tool-aware config wiring + leak regressions, build.sh dispatch ↔ registry/default-set sync, protected-path defaults + placeholder materialization.
- **Integration (Go, real container, skipped without Incus + coi-default)** — `internal/session/setup_codex_integration_test.go`: seeding lands byte-identical (comments/table-headers in the TOML prove it was copied, not parsed), correct ownership, no synthesized state file; AGENTS.md managed block coexists with host content and stays a single copy across sessions.
- **Integration (Python)** — `tests/config/test_codex_config_seeding.py` (end-to-end `coi shell` with `COI_USE_DUMMY=1`: verbatim `config.toml`, seeded `auth.json`, exactly one COI block in `~/.codex/AGENTS.md`), `tests/mount/test_codex_config_protection.py` (read-only by default, planting-when-absent blocked, untrusted `writable_paths` opt-out ignored — mirrors the claude settings protection suite), plus codex cases in `tests/build/test_agent_selection.py`.

`go build ./... && go vet ./... && go test ./...` green; `ruff check` / `ruff format --check` green on `tests/`.

## Not verified here (needs a real Incus host)

The live-container smoke: `[container.build] agents = ["claude", "codex"]` + `coi build --force`, then `[tool] name = "codex"` + `coi shell` → direct codex launch in bypass mode, seeded auth (`codex login status`), managed block in `~/.codex/AGENTS.md`, and a `--resume` round-trip. Residual risks are upstream flag drift and whether `codex resume` accepts the mode flags — both degrade gracefully and are confined to `BuildCommand`. @breml offered to test on their setup in #698.


mensfeld:
An adversarial multi-angle self-review of this PR (8 finder angles + per-candidate verification) surfaced several confirmed issues, all fixed in a8b3f90:

**Correctness**
- **Shell injection via `[tool.codex]` launch values** (worst finding): `model`/`reasoning_effort` travel as launch flags through a shell command string (`strings.Join` → `bash -c`, nested again in the tmux path), and the `[tool]` section is mergeable from an *untrusted* project `.coi/config.toml` — a cloned repo could set `model = "$(…)"` and execute shell inside the container at launch. Values are now validated as single shell-safe tokens: `getConfiguredTool` rejects bad values with a loud error, and the setters drop them fail-closed. (Claude is unaffected — its model rides an env var.)
- **Resume could pin the wrong session**: `DiscoverSessionID` picked the newest rollout by mtime, but `incus file pull` does not preserve mtimes, so after a save/restore round-trip "newest" degenerated to "last file pulled". Discovery is now path-ordered (the layout is date+timestamp encoded), with a regression test asserting mtimes are ignored.
- **`install_codex` retries never fired**: the `curl | sh` pipeline's exit status was `sh`'s, and `sh` exits 0 on the empty stdin a failed `curl` leaves behind — a transient network failure "succeeded" and hard-failed the build at the binary check. Fixed with `set -o pipefail` inside the `su` login shell. Note: `install_claude_cli` and `install_pi` share this pre-existing flaw (their pipelines also run inside `su -c`); left for a follow-up to keep this PR scoped.
- **The registry↔build.sh default-set sync test compared only counts** — a same-length drift (default gaining codex, dropping pi) would have passed while the footgun warning reasoned about the wrong set. It now compares the actual dispatched installer lists (membership + order).

**Cleanup**
- gai.conf IPv4 guard hoisted to `prefer_ipv4()` called from `main()` — opencode/pi installers (which had no guard) now benefit too.
- `_run`/`_make_workspace_writable` hoisted into `tests/support/helpers.py` and all six private copies (this PR's new file was the 4th/5th) switched to imports, per the #684 convention.
- `fakeCodexSetupManager` now embeds the existing `fakeAutoCtxManager`; dead `//nolint:nilerr` directives removed; five stale "installs all supported agents" comments left behind by the opt-in change fixed.

**Reviewed but deliberately unchanged**
- `codex resume <uuid>` + root flags: current codex docs accept them; the assumption and its degrade path are documented at the single fix site in `BuildCommand`.
- The `.codex/config.toml` read-only placeholder is created in every workspace regardless of tool — this mirrors the existing `.claude/settings.json` behavior from #504 (the planting threat is tool-independent), but it does mean claude-only users will see an untracked `.codex/` in `git status`; flagging in case you'd rather gate it.
- Verification also refuted one suspicion: `codex resume --last` with no sessions starts fresh upstream (no failed-launch risk). The inverse edge (host discovery finds a UUID but in-container restore failed → `resume <uuid>` exits with a missing-session error) is real but rare and already preceded by a loud restore warning.

mensfeld:
Follow-up folded in after all: e23f39c applies the same `set -o pipefail` retry fix to `install_claude_cli` and `install_pi` — their `curl | bash`/`curl | sh` pipelines inside `su -c` had the identical swallowed-curl-failure flaw as the codex installer (verified empirically: the old pattern exits 0 on an unresolvable host, the fixed one propagates curl's exit 6). `install_opencode` was already correct since its pipeline runs in the outer script where `pipefail` applies.

breml:
@mensfeld I quickly wanted to test this feature just now, but I failed, since I don't have package `systemd-devel` available and I can not easily install it, since I am running on an immutable Fedora (Bluefin). Therefore I need to create a dev vm later on for this, which takes a bit more time and I have to see, when I can squeeze this in.

breml:
@mensfeld I did a quick "smoke" test and this worked without issues. Specifically, I changed `tool.name = "codex"` and I already had `tool.permission_mode = "bypass"` set. I started my container built earlier (which already included codex based on the build.sh snipped I shared in the issue) and `coi shell` happily started the container and jumped right away into codex. After authentication with device token, I had codex running in YOLO mode (as expected).

One thing that I kind of miss is a way to start directly into one tool or the other. Basically, I would like to have `tool.name = "claude"` (since this is my default), which allows me to use `coi shell`. But then I would like to also have `coi shell --codex` (similar to `coi attach --bash`) or even `coi attach --codex`.

mensfeld:
Thank you. I will merge and release it then.

About your second request let me think about it from the UX perspective for a moment. I will spin up a separate card for this.

mensfeld:
ref https://github.com/mensfeld/code-on-incus/issues/708

PR #704, opened 2026-08-14
feat(network): DNS resolver pinning and egress port allowlist

mensfeld:
## Summary

Two composable, **trusted-scope-only** egress controls that shrink a compromised container's blast radius (e.g. a malicious dependency that gets RCE and starts probing the LAN). Both are no-ops when unset — with neither key set, emitted nft rules are byte-identical to before.

### `[network] dns_servers` — pin the resolver (restricted mode)
```toml
[network]
mode = "restricted"
dns_servers = ["192.168.1.2"]   # e.g. your Pi-hole
```
Accepts `:53` only to the listed IPv4 addresses and rejects all other off-box DNS, so a compromised container cannot bypass your resolver by talking straight to `8.8.8.8` or a hardcoded one. Implemented **forward-chain only**, so the bridge's own resolver keeps working with no `resolv.conf` changes; a pinned LAN resolver stays reachable on `:53` even when `block_private_networks` is on (ordered before the RFC1918 block). **Errors out in allowlist mode**, which blocks all DNS by design — re-opening `:53` there would reintroduce the exact host/container divergence that mode prevents.

### `[network] allowed_ports` — cap outbound ports
```toml
[network]
mode = "restricted"
allowed_ports = [80, 443]       # web only; blocks SSH, DB ports, IoT panels, ...
```
In restricted mode this flips allow-by-default into "internet, but only these dports" (ICMP echo still permitted, so ping/health checks work); in allowlist mode it further constrains the allowlisted hosts. Bridge-provided DNS is unaffected — add `53` if the container resolves via an off-box resolver.

The two **compose**: a pinned resolver is reachable on `:53` regardless of `allowed_ports`, giving the full "use my Pi-hole, on these ports only" posture.

## Security notes
- Both keys are stripped from untrusted project configs in `sanitizeUntrustedNetwork` (a resolver pin is a DNS-redirect primitive; the port cap is honored from trusted scope for a uniform rule). Merged, cloned, and added to the `additionalProperties:false` JSON schema.
- Validation is **fail-closed**: a bad `dns_servers`/`allowed_ports` value aborts setup with the boot block still in place.
- **Caveat (documented):** DNS pinning only bites when 443 is also constrained (allowlist mode or `allowed_ports`) — otherwise malware can still tunnel DNS-over-HTTPS on 443.

## Testing
- **Unit:** validators, port-set token builders, a historic-parity test pinning that no-ports output is byte-identical to the old hard-coded clause, config merge, and untrusted-stripping.
- **Integration** (`egress_integration_test.go`): allowlist port cap, combined restricted DNS-pin + port cap, and the allowlist/`dns_servers` incompatibility guard.
- Emitted `nft` rules were smoke-tested against a real kernel — every rule (`th dport { 80, 443 }`, single-port, DNS pin accept/reject, rate-limited ICMP) is accepted as-is.

## Scope
This is Phases 1 & 2 of a 3-phase plan. **Phase 3** — per-destination `host:port` syntax in `allowed_domains` (the port-bucket set refactor + rule-ordering change so specific LAN services can be allowlisted on a single port) — is a follow-up.


mensfeld:
Added a comprehensive test commit (`6f1211b`) on top — ~35 new egress-specific tests across four layers:

- **Unit:** validator edge cases (dedup/sort/trim, IPv4-only, boundaries, error-message specificity, input-not-mutated), token-builder nft syntax, nil==empty==historic parity, and fail-closed `ApplyRestricted`/`ApplyAllowlist` paths (reject bad `dns_servers`/`allowed_ports` before any nft call) — all run without Incus.
- **Config:** TOML round-trip, merge replace-vs-nil semantics, profile inheritance, `synthesizeDefaultProfile` deep-copy (aliasing guard), and decode+sanitize of an untrusted project TOML.
- **Schema:** validation accepts the keys, enforces the 1–65535 port bound (rejects 0 / 70000 / non-int / non-string), and still rejects unknown network keys.
- **Integration + E2E (network job):** rule-assertion tests (multi-resolver pins, single-port normalisation, historic-parity regression, static+dynamic set scoping, and the security-critical ordering: pinned LAN resolver `:53` accept precedes the RFC1918 block), plus **true end-to-end tests that drive real TCP connections from inside the container** and confirm an allowed port connects while a blocked port is rejected, a pinned resolver answers on `:53` while an unpinned one is blocked (non-53 unaffected), and an allowlisted host is constrained to its permitted port. Each E2E test baseline-checks connectivity and skips (never false-passes) when the runner has no egress.

Assertion predicates were validated against real `nft` output on the kernel. Full local suite (`go test ./...`) is green.

mensfeld:
**Important coverage note + fix (`6dbc065`).**

While verifying CI I found that the **network lane runs the Python E2E suite (`tests/network/*.py`), not the Go `*_integration_test.go` files** — those self-skip without Incus in the unit lane and aren't executed anywhere with Incus. So the Go integration/E2E tests I added give local + compile-time value but **weren't actually enforcing the feature in CI**.

Added `tests/network/test_egress_dns_and_ports.py` — real end-to-end tests that run in the network lane against real Incus + nftables + internet, driving TCP connections from inside the container (bash `/dev/tcp` + curl):

- **restricted + `allowed_ports=[443]`** — 443 reachable, **80 and 22 rejected** (same host, so the block is attributable to the port).
- **restricted + `dns_servers=[1.1.1.1]`** — `:53` to the pinned resolver allowed, `:53` to an unpinned resolver **rejected**, non-53 unaffected, and normal DNS resolution still works (bridge resolver untouched).
- **allowlist + `allowed_ports=[443]`** — allowlisted host reachable only on 443, **blocked on 80**, non-allowlisted host **blocked entirely**.
- **fail-closed** — `dns_servers` in allowlist mode is refused; an invalid `dns_servers` IP aborts the session and never runs the command.

These use `write_trusted_coi_config` (COI_CONFIG) because `dns_servers`/`allowed_ports` are trusted-scope only — a workspace config would strip them. Each "blocked" assertion is paired with an "allowed" one so it can't pass for the wrong reason. `ruff check` + `ruff format --check` over `tests/` are clean. Watching this run now.

mensfeld:
### CI result: feature is green; one unrelated flake retriggered

**All feature coverage passed** on run `31798479887`:
- **Integration Tests (network): pass (22m31s)** — all 5 new Python E2E tests ran against real Incus and passed:
  - `test_restricted_allowed_ports_blocks_other_ports` ✅ (443 reachable, 80/22 rejected)
  - `test_restricted_dns_pin_blocks_unpinned_resolver` ✅ (pinned :53 allowed, unpinned :53 blocked, non-53 unaffected, bridge DNS works)
  - `test_allowlist_allowed_ports_constrains_host` ✅
  - `test_invalid_dns_server_fails_closed` ✅
  - `test_dns_servers_rejected_in_allowlist_mode` ✅
- **Unit Tests: pass** — the Go unit/config/schema coverage.

**Sole failure** was `monitoring-advanced`:
`test_security_monitoring.py::TestAuthLogWatcher::test_sudo_not_in_sudoers_triggers_high` (`AssertionError: got events: []` after 100s).

This is **unrelated to this PR** and a known flake class:
- This branch touches only network-egress code + tests; it has zero effect on the auth-log-watcher.
- `monitoring-advanced` **passed** on the immediately prior commit (`84fe8ae`, 9m10s) with identical monitoring code — the only change since was adding a `tests/network/*.py` file.
- The repo has a documented history of monitoring timing flakes cleared by retrigger (`bb08d22`, `9d4e108`, `d2ecce8`).

Pushed an empty `ci: retrigger` commit (`0afe036`) to resample, matching that convention. My token can't re-run jobs directly. Watching the new run.

mensfeld:
### Fixed review findings 1 & 2 (`ebe1645`)

Root cause: `allow_local_network_access` emitted an **all-ports blanket accept** for the RFC1918 ranges, which sat before the port-cap default reject — so with `allow_local` + `allowed_ports=[443]` the LAN stayed reachable on *every* port (SSH/DB/admin panels), the exact lateral-movement surface the cap exists to close (finding 1). It was also inconsistent with `dns_servers`, which already filters `:53` on the LAN (finding 2).

Fix — make both controls apply uniformly to LAN traffic:
- New `addLocalNetworkAllows` helper **port-scopes the RFC1918 accept to `allowed_ports`** when a cap is set (shared by restricted + allowlist mode). With no cap it stays the historic all-ports blanket accept, so configs without `allowed_ports` emit **byte-identical** rules (parity test added).
- `dns_servers` already restricts `:53` everywhere (pin reject precedes the RFC1918 allow); now documented that a LAN resolver must be listed by exact IP to stay reachable — so the two controls are consistent.
- Precedence documented in config comments, README, starter TOML, and JSON schema.

Integration tests (network lane):
- `allow_local` + `allowed_ports=[443]` → every RFC1918 allow rule is port-scoped.
- `allow_local`, no cap → RFC1918 allow rules stay blanket (parity guard).
- `allow_local` + `dns_servers` → order is pinned-LAN-resolver `:53` accept < `:53` reject < LAN allow, so the pin filters LAN `:53` while the pinned resolver stays reachable.

No clean Python E2E is possible here (no non-gateway LAN target exists in CI; the gateway is exempt via its own all-ports rule), so this is covered by rule-assertion integration tests — matching how the repo tests rule structure. Rule syntax (CIDR daddr + dport set) verified against a real kernel; unit/config/schema suites green.

(Note: the earlier `monitoring-response` flake cleared on re-run — `CI Success` passed on `0afe036` before these fix commits.)

mensfeld:
### CI stabilization (`d7cb844`) + CI-enforced fix tests (`60a0be8`)

**Stabilization** — the integration lanes have a documented tail of timing flakes (monitoring `test_threat_deduplication`, auth-log/inotify watchers, shell-ephemeral poweroff — #653/#658/#665) with no retry mechanism, so one transient miss reds the whole run. Added `pytest-rerunfailures` and `--reruns 2 --reruns-delay 3` to the shared integration pytest step:
- A transient flake is resampled automatically (proven locally: a first-attempt failure passes on rerun).
- A genuine bug still fails after every attempt.
- **Integration lanes only** — the Go unit lane stays strict (no reruns), so real regressions are never masked.
- Since PRs run their own workflow, this stabilizes *this* run immediately.

**Fix tests** — the Go rule-assertion tests for the LAN port-cap fix run only locally (network lane = Python). Added Python rule-assertion tests that query `ip coi forward` and assert the RFC1918 LAN allow rules are port-scoped under a cap, blanket without one — so the fix is CI-enforced.

This run is the first with retries active; expecting the previously-flaky lanes to self-resample green.

mensfeld:
### Stabilization is working — two layers now covered

**Layer 1 (test-timing flakes) — fixed.** On the run with `pytest-rerunfailures` active, both previously-red lanes went green: **`monitoring-response` pass**, **`shell-ephemeral` pass**, plus **`network` pass**. That was the dominant recurring red.

**Layer 2 (image-build infra flake) — was the sole remaining red.** `monitoring-nft` failed at the *Build COI image* step (before pytest, so reruns can't help) with `curl: (35) Recv failure: Connection reset by peer` — a transient network reset in the build script that exhausted the old 3-attempt/10s retry (~20s window). Widened it to **5 attempts / progressive backoff (~150s window)** in `d56451b` so a brief mirror blip self-heals.

Both flake layers are now hedged. Watching the new run.

mensfeld:
### Two changes per request (`fcfd333`, `cf175ac`)

**1. Removed the chronically-flaky `test_threat_deduplication`** (`fcfd333`). It's a sticky timing flake (the "past the 30s dedup window, alert should re-fire" assertion) that fails *all* same-process reruns within a job (166s each), so `pytest-rerunfailures` couldn't rescue it — and it was the dominant CI red. Tracked in #653/#658/#665; removed rather than quarantined, as requested.

**2. Surfaced the egress limits in the agent context** (`cf175ac`). `SANDBOX_CONTEXT.md` (also injected into the tool's system prompt via the auto-context file) previously described only the network *mode*, so an agent under `allowed_ports`/`dns_servers`/allowlist didn't know which ports/resolvers were blocked and would waste turns dialing them. Now `RenderContextFileContent` appends:
- the destination-port allowlist ("outbound is restricted to destination port(s) …"),
- the pinned DNS resolver(s) ("DNS is pinned to … on port 53"),
- (allowlist mode) the enumerated reachable destinations.

The three fields are threaded from `NetworkConfig` → `ContextInfo` in session setup. No controls set ⇒ context renders unchanged (parity test). Tests: Go unit tests per-control + combined + parity, and a **network-lane E2E** that reads the rendered `SANDBOX_CONTEXT.md` inside the container and asserts the port cap + pinned resolver appear.

Local gate before push: `ruff` + `ruff format` clean, `golangci-lint run ./...` 0 issues, `gofmt` clean, all touched Go suites green.

PR #706, opened 2026-08-14
fix(image): mask udisks2 so coi containers don't break host suspend

blegat:
I noticed that when closing my laptop and opening it back, I needed ~15 seconds before it was accepting any input.
Looking into this, it turns out that closing the lid never actually sleeps the laptop.
In the log (Omarchy host with Linux kernel 7.1.4), I saw
```
Freezing user space processes failed after 20.008 seconds
(1 tasks refusing to freeze, wq_busy=0):
task:udisksd  state:D  ... fuse_file_poll ... do_sys_poll
```

It turns out this is due to the udisks2 process of the container (an fwupd dependency in the Ubuntu base image). It keeps `/proc/swaps` open to watch for swap changes. Incus backs `/proc/swaps` with the `lxcfs` on the host.

When the host suspends, the kernel freezer stops lxcfs so udisks doesn't get any answer anymore, it is stuck in D state and cannot be frozen. The freezer gives up after its 20s timeout, the suspend aborts with "Device or resource busy", and logind immediately retries, forever.

I tried applying the change of this PR locally and it fixes the issue for me.


mensfeld (approved):

mensfeld:
Great debug. Thank you for this PR and sorry for the trouble.

PR #710, opened 2026-08-18
feat(network): per-host ports on [[network.hosts]] (Phase 4)

mensfeld:
## What

A `[[network.hosts]]` entry can now carry its own `ports`, scoping the firewall reachability of that **one host** without touching the rest of egress. This is the missing piece that lets **restricted** mode open a single LAN service on a single port while the internet stays fully open — which the global `allowed_ports` can't do, since it caps every destination (internet included).

The motivating posture — *"internet open, on the LAN reach only `redmine.susanoo.pl:443`, and my Pi-hole resolves it to its local address"*:

```toml
[network]
mode        = "restricted"          # all internet open; LAN + metadata blocked by default
dns_servers = ["192.168.1.2"]       # Pi-hole (reachable on :53 even though the LAN is blocked)

[[network.hosts]]
ip        = "192.168.1.50"          # redmine's LAN address
hostnames = ["redmine.susanoo.pl"]  # → /etc/hosts, so it resolves to the local IP
ports     = [443]                   # ...and reachable ONLY on 443
```

## How

- **config:** `HostEntry` gains `Ports []int`; `ValidateNetworkHosts` checks `1..65535` (at config load and via `coi hosts add`).
- **`applyHostFirewall`:** resolves the effective ports (entry's own → global `allowed_ports` → all) and feeds them to the restricted-mode targeted allow (`insertContainerAcceptRule`) and the allowlist-mode port-scoped set tuple (`AddStaticTuples`).
- **CLI:** `coi hosts add <container> <ip> <hostname> --ports 443`.
- **docs:** JSON schema, `config.starter.toml`, README (with the redmine recipe), CHANGELOG (0.12.0).

## Compatibility

Empty `ports` keeps the prior behavior (inherit `allowed_ports`, else all ports), so existing `[[network.hosts]]` configs are byte-unchanged. Trusted-scope only, like the rest of `[[network.hosts]]`; applied in restricted + allowlist mode, ignored in open mode.

## Tests

- **Config unit:** `ValidateNetworkHosts` accepts valid ports and rejects 0 / >65535 / negative / a bad value in a list.
- **Real-nft integration** (`TestHostFirewallPerHostPorts`): a restricted-mode entry's targeted allow is scoped to the entry's own ports; entry ports **override** the global cap; a portless entry falls back to it.
- **End-to-end network lane** (`test_restricted_host_entry_per_entry_ports_decouple`): with no `allowed_ports`, a `ports=[443]` host entry is scoped to 443 **while the container still has a blanket all-ports internet accept** — proving the decoupling.

Verified locally with golangci-lint v2.12.2 (0 issues), `go test`, and `ruff` — all clean.

Completes the egress-hardening arc (dns_servers #704, allowed_domains per-destination ports #709, and now per-host ports) so a compromised container can be boxed in by resolver, ports, and per-destination — including a single pinhole to one LAN service.


mensfeld:
Addressed all four Copilot remarks in f996444:
- **hosts.go** — `GetIntSlice("ports")` error is now handled/returned.
- **user_hosts.go** — a per-host port validation error is now attributed to `network.hosts <ip>` instead of the misleading `allowed_ports:` wording (both inputs are pre-validated, so the path is practically unreachable, but the message is now correct).
- **nftset_integration_test.go** & **test_egress_dns_and_ports.py** — port assertions now use a word-boundary regex (`\b443\b`) so a rule scoped to `8443` no longer satisfies a check for `443`.

Go build/lint/test + ruff all green locally.

PR #713, opened 2026-08-22
fix(installer): detect Incus init by MANAGED network, not any network (#703)

### LINKED ISSUE #703: setup script - ensure_incus_initialized() never runs incus admin init on real hosts (unmanaged interfaces pollute the network-list guard)
Note: this situation is probably only because i never used incus before.

After setup script, i've tried (like recommended):
` incus profile device set default root pool=btrfs-pool` (my system has no zfs)
but incus responded with:
`Error: Device doesn't exist`

i gave glm-5.2 the error to analyse the setup script and the root cause of the problem:

## Summary

On any real host with a physical NIC and loopback, the installer's
`ensure_incus_initialized()` guard is always true, so `incus admin init --auto`
is **never run**. As a result the `default` profile is left with **no devices**
(no `root` disk, no `eth0` nic) and `incusbr0` is never created. The subsequent
`incus profile device set default root pool=btrfs-pool` then fails with
`Error: Device doesn't exist`, and the script's printed "manual recovery" command
fails identically because it also uses `set` instead of `add`.

## Environment

- OS: Arch Linux (Omarchy desktop, Hyprland)
- Incus: `incus 7.2.0-1` (pacman)
- `code-on-incus` installer from `master` (`install.sh`)

## Symptom

After install, the `default` profile is completely empty:

```
$ incus profile show default
config: {}
description: Default Incus profile
devices: {}          # ← no root, no eth0
name: default
used_by: []
```

No managed bridge exists:

```
$ incus network list
+--------+----------+---------+ ...
|  NAME  |   TYPE   | MANAGED | ...
+--------+----------+---------+ ...
| enp7s0 | physical |    NO   | ...
| lo     | loopback |    NO   | ...
+--------+----------+---------+ ...
```

And the line the script tells the user to run manually also fails:

```
$ incus profile device set default root pool=btrfs-pool
Error: Device doesn't exist
```

## Root cause

`ensure_incus_initialized()` (around lines 455–486 of `install.sh`) decides
whether Incus is initialized by checking if `incus network list --format=csv`
returns *anything*:

```bash
local networks
if ! networks="$(incus network list --format=csv 2>/dev/null)"; then
    ...
    return 1
fi
if [ -n "$networks" ]; then
    return          # ← skips `incus admin init --auto`
fi
```

The comment says *"an empty list means never initialized."* But
`incus network list` includes **unmanaged** physical and loopback interfaces,
which are present on every real host. So `$networks` is never empty in practice
and `incus admin init --auto` is effectively always skipped. The guard only
behaves as intended inside a pristine VM/container that has no interfaces at all
(presumably how it was tested).

I confirmed this empirically — the CSV output is:

```
enp7s0,physical,NO,,,,0,
lo,loopback,NO,,,,0,
```

i.e. the list is non-empty purely because of unmanaged interfaces, so the guard
returns early before init runs. Because init is skipped, the `default` profile
never gets the `root`/`eth0` devices that the later `profile device set root`
call assumes already exist.

A secondary issue: when the `set` fails, the script prints a "manual recovery"
hint (lines ~651–652):

```bash
echo -e "${YELLOW}  You can manually configure it later with:${NC}"
echo -e "  ${BLUE}incus profile device set default root pool=btrfs-pool${NC}"
```

That suggestion is the wrong verb for an empty profile — it must be `add`, not
`set` — so following the script's own advice fails the same way.

## Expected vs actual

- **Expected:** installer runs `incus admin init --auto` (creating `incusbr0`
  and the `root`/`eth0` devices on the `default` profile), then repoints `root`
  to the btrfs pool.
- **Actual:** init is skipped due to the faulty guard; profile stays empty;
  the btrfs pool is created but can't be attached; the printed manual command
  also fails.

## Suggested fix

Filter on the **managed** column (field 3 of the CSV) instead of any output:

```bash
    local networks
    if ! networks="$(incus network list --format=csv 2>/dev/null)"; then
        echo -e "${YELLOW}⚠ Unable to determine whether Incus has been initialized${NC}"
        echo "  Could not query Incus networks. Ensure the Incus daemon is running and your user has access."
        return 1
    fi
    # Only a MANAGED network means `incus admin init` has run. Unmanaged
    # physical/loopback interfaces are present on every real host and would
    # otherwise make this non-empty, falsely suppressing init.
    if printf '%s\n' "$networks" | cut -d',' -f3 | grep -q '^YES$'; then
        return
    fi
```

(Verified the CSV column order: `name,type,managed,ipv4,ipv6,description,used_by,state`.)

And for the recovery message, either point users at `add` or, better, make the
script tolerant of both states by adding the device if `set` reports it missing.

## Workaround for affected users (example with btrfs)

```bash
incus profile device add default root disk pool=btrfs-pool path=/
incus network create incusbr0
incus profile device add default eth0 nic network=incusbr0
```

AmirF194:
Root cause: `ensure_incus_initialized()` decides whether Incus has been set up by checking if `incus network list --format=csv` returns any output at all. On a real host that list always includes unmanaged physical and loopback interfaces (confirmed against Incus's own CLI source: default columns are `name,type,managed,ipv4,...`, MANAGED is `YES`/`NO`), so the list is never empty and `incus admin init --auto` is skipped even on a fresh install. The default profile is then left deviceless, and the next step's `incus profile device set default root pool=...` fails with "Device doesn't exist" - as does the manual recovery command the script prints for that failure, which used the same `set` instead of `add`.

Fix: filter the CSV on the MANAGED column (field 3) instead of on any output, and point the two recovery hints at `incus profile device add ... disk pool=... path=/` instead of `set`.

Verification:
- New Go test `TestInstallSh_EnsureIncusInitialized_RunsInitWhenOnlyUnmanagedNetworksExist`, stubbing `incus network list` with the exact CSV shape from the report (unmanaged physical NIC + loopback): fails on current master (init is skipped), passes on this branch.
- Existing `TestInstallSh_EnsureIncusInitialized_SkipsWhenAlreadyInitialized` updated to stub a genuinely managed network row, since its old stub had an empty MANAGED field and would no longer represent "already initialized" under the new check.
- Ran the full `internal/installer` package test suite in a clean Docker container (`golang:1.25`, with `libsystemd-dev` for the build): all pass except `TestInstallSh_NftAutoSetup_NonInteractive`, which also fails identically on unpatched master in the same container (it needs outbound `apt-get install nftables`, unavailable here) and is unrelated to this change.
- `go build ./...` and `go vet ./...` clean; `shellcheck install.sh` shows only the two pre-existing SC2064 warnings at lines 275/324, unchanged by this diff.
- Did not verify against a real Incus daemon; the fix and both new/updated tests are shell/CSV-parsing logic exercised through stubbed `incus`/`sudo` binaries, the same pattern the existing test suite already uses for this function.

Fixes #703

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -458,8 +458,9 @@ ensure_incus_initialized() {
         return
     fi
 
-    # Check if Incus has been initialized by looking for any networks.
-    # `incus admin init` creates incusbr0; an empty list means never initialized.
+    # Check for a MANAGED network (CSV column 3), not just any network: unmanaged
+    # physical/loopback interfaces are on every real host and would otherwise
+    # make this non-empty, falsely skipping init (#703).
     # If the query itself fails (daemon down, no permissions), warn and bail out
     # rather than incorrectly triggering init.
     local networks
@@ -468,7 +469,7 @@ ensure_incus_initialized() {
         echo "  Could not query Incus networks. Ensure the Incus daemon is running and your user has access."
         return 1
     fi
-    if [ -n "$networks" ]; then
+    if printf '%s\n' "$networks" | cut -d',' -f3 | grep -q '^YES$'; then
         return
     fi
 
@@ -589,7 +590,7 @@ setup_zfs_storage() {
                 printf "${YELLOW}  %s${NC}\n" "$profile_output"
             fi
             echo -e "${YELLOW}  You can manually configure it later with:${NC}"
-            echo -e "  ${BLUE}incus profile device set default root pool=zfs-pool${NC}"
+            echo -e "  ${BLUE}incus profile device add default root disk pool=zfs-pool path=/${NC}"
         fi
     else
         echo -e "${YELLOW}⚠ ZFS storage pool creation failed${NC}"
@@ -649,7 +650,7 @@ setup_btrfs_storage() {
                 printf "${YELLOW}  %s${NC}\n" "$profile_output"
             fi
             echo -e "${YELLOW}  You can manually configure it later with:${NC}"
-            echo -e "  ${BLUE}incus profile device set default root pool=btrfs-pool${NC}"
+            echo -e "  ${BLUE}incus profile device add default root disk pool=btrfs-pool path=/${NC}"
         fi
     else
         echo -e "${YELLOW}⚠ btrfs storage pool creation failed${NC}"
--- a/install.sh
+++ b/install.sh
@@ -458,18 +458,26 @@ ensure_incus_initialized() {
         return
     fi
 
-    # Check for a MANAGED network (CSV column 3), not just any network: unmanaged
-    # physical/loopback interfaces are on every real host and would otherwise
-    # make this non-empty, falsely skipping init (#703).
-    # If the query itself fails (daemon down, no permissions), warn and bail out
-    # rather than incorrectly triggering init.
-    local networks
+    # Decide whether Incus has been initialized without being fooled by the
+    # unmanaged physical/loopback interfaces that appear on every real host: a
+    # bare "is the network list non-empty?" check is never empty and falsely
+    # skips init (#703). `incus admin init --auto` creates both a MANAGED network
+    # (incusbr0) and a storage pool, so treat either as proof of initialization -
+    # a host set up with an existing/custom network and no managed bridge still
+    # has a pool, which is the reliable signal.
+    # If the network query itself fails (daemon down, no permissions), warn and
+    # bail out rather than incorrectly triggering init.
+    local networks pools
     if ! networks="$(incus network list --format=csv 2>/dev/null)"; then
         echo -e "${YELLOW}⚠ Unable to determine whether Incus has been initialized${NC}"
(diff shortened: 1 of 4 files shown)


mensfeld:
Good catch. Thank you. Will pick it up tomorrow :pray:

AmirF194:
Sounds good, no rush at all. Let me know if the review turns up anything else.

mensfeld:
Added small remarks.

mensfeld (approved):

mensfeld:
@AmirF194 if you are ok with my changes I will merge.

AmirF194:
Yes, very happy with these. The storage-pool fallback closes a real gap in my original fix (a custom-network host with no managed bridge), and the real-daemon meta test is a good addition since the Go tests can only assert an assumed CSV shape. Please go ahead and merge.

mensfeld:
Great. Thank you. I should craft a release in the upcoming days tops.

AmirF194:
Thanks for the quick review and for tightening up the fallback and the meta test. Good luck with the release.

PR #719, opened 2026-08-24
fix(health): flag non-thin LVM pools in coi health (#686)

### LINKED ISSUE #686: coi health: non-thin LVM pools (lvm.use_thinpool=false) behave like dir but aren't flagged

AmirF194:
`evaluatePool` flags a pool as a warning when every launch does a full copy instead of a CoW clone. It only checked for `driver == "dir"`, so an `lvm` pool with `lvm.use_thinpool` disabled, and any `lvmcluster` pool, reported healthy even though both do a full logical-volume copy on every launch for the same reason `dir` does.

`lvmcluster` never gets a thin pool (Incus disables it unconditionally for clustered LVM); a plain `lvm` pool only loses its thin pool when `lvm.use_thinpool` is explicitly set to a falsy value, since Incus defaults an absent key to true. Both cases mirror the driver's own `usesThinpool()` logic (`internal/server/storage/drivers/driver_lvm_utils.go` upstream).

The `#684 follow-up` PR deferred this exact case, on the assumption that detecting it needed an extra `incus storage show` call per pool. That is not the case: `incus storage list --format=json` already returns each pool's full `config` map in the same response `listPoolDrivers` reads for the driver name, so this reads that same call's output instead of adding a new one.

`isNonThinLVM(driver, config)` classifies a pool from that config; `evaluatePool` now downgrades an OK status to a warning, and appends the explanatory message, for `driver == "dir"` or a non-thin LVM pool alike, on both the usage-ok and usage-error paths (matching the existing schema-consistency guarantee for `dir`).

Verification:
- `go test -race ./internal/health/... ./internal/container/...`: all green, including the new `TestIsNonThinLVM` and `TestCheckIncusStoragePools_NonThinLVM*` cases (a non-thin `lvm`/`lvmcluster` pool with healthy usage now reports `warning`, not `ok`).
- `golangci-lint run`, `go build ./...`, `go vet ./...`, `gofmt -l .`: clean.
- Reproduced the reported symptom directly: called the real `CheckIncusStoragePools` against a fake `incus` binary answering `storage list`/`storage info` for an `lvm` pool with `lvm.use_thinpool: "false"` and healthy usage. It reported `ok` before this change and `warning` after, with no other test doubles involved.
- Did not verify against a real Incus daemon backed by an actual LVM volume group (no such host available here); the fake-`incus` repro and the config-shape check above are the only coverage.

Fixes #686


mensfeld:
Yet another edge case :sweat:

mensfeld:
I will do my best to tackle this one today :pray:

mensfeld:
Thank you once more :pray:

AmirF194:
Glad it's sorted, thanks for the quick turnaround on reviewing it.

Issue #726, opened 2026-08-26
coi shell ignores [container] storage_pool — only coi run/coi build honor it


technicalpickles:
## Summary

`[container] storage_pool` in config.toml is only honored by `coi run` and `coi build`. `coi shell`, the primary interactive command, ignores it completely and always launches on whichever pool the Incus `default` profile points at.

`internal/session/setup.go`'s `SetupOptions` struct (line 30) has no `StoragePool` field, and `Setup()` (line 92) never references one. Compare `internal/cli/phases_run.go`, which reads `a.cfg.Container.StoragePool` and passes it into `launchOrReuseContainer` (line 291, defined in `internal/cli/run.go:177`). `internal/cli/phases_shell.go`'s `configureSessionPhase` builds a `session.SetupOptions{...}` (line 363) with the same config object in scope, but there's no field to put `StoragePool` into, so it's silently dropped.

## How I found this

Host: OrbStack VM, aarch64, Incus 7.2. ZFS can't work on OrbStack (no zfs.ko for the injected kernel), so I set up a `btrfs-pool` alongside the default `dir` pool and pointed `storage_pool = "btrfs-pool"` at it in config.toml, expecting new containers to land there. `incus storage list` showed images cached on `btrfs-pool`, but every actual `coi-*` container instance was still on `dir`, including ones created same-day, well after the config change.

Bisected with a throwaway workspace:

1. `coi run -- sleep 8` from the workspace -> `incus list -c n,b` shows the container on `btrfs-pool`. Config is being read correctly, at least somewhere.
2. `coi shell` in the same workspace, same config, same profile -> container lands on `dir`.
3. Traced both call paths. `phases_run.go` passes `a.cfg.Container.StoragePool` through explicitly. `phases_shell.go` never reads `Container.StoragePool` at all, and `session.SetupOptions` has nowhere to put it if it did.

Confirmed still current on `master` after pulling today (`v0.11.2-25-gadb2e8d`): grepped `StoragePool` across `internal/session/setup.go`, `internal/session/types.go`, and `internal/cli/phases_shell.go`, zero hits.

## Impact

Anyone setting `storage_pool` expecting it to apply to their actual interactive sessions (which is what `coi shell` is for) gets silently routed to the default pool instead. No error, no warning, just slower container creation than expected with no indication why.

## Suggested fix

[...]

Issue #373, opened 2026-05-14
coi run ignores [network] and [ssh] config settings (coi shell ok)


mensfeld:
coi run applies limits, mounts, security paths, and timezone from config, but silently skips [network] and [ssh] settings. A container started with coi run always has unrestricted network access and no SSH agent forwarding,
  even when the config explicitly sets network.mode = "restricted" or ssh.forward_agent = true.

PR #729, opened 2026-08-26
fix(run): apply shell-parity container hardening/setup in coi run (#726 follow-up)

mensfeld:
Follow-up to #726/#727. The audit for the `storage_pool` drop found that `coi run` and `coi shell` are **two separate launch paths** (shell uses `session.Setup()`; run hand-rolls `phases_run.go`), and **6 more settings the shell path applies were silently dropped by `coi run`** — mostly leaving `coi run` *less hardened* than `coi shell` for the same config. This threads each into the run pipeline at the same point shell applies it, with an end-to-end test per fix.

## Fixes (ranked by impact)

| # | Setting | Impact |
|---|---------|--------|
| 1 | **NIC anti-spoofing** (`EnableNICSecurity`: eth0 `security.ipv4_filtering`/`mac_filtering`/`port_isolation`) | **HIGH** — without it a restricted/allowlisted `coi run` can spoof its source IP/MAC to dodge the `saddr`-keyed nft egress allowlist, and reach sibling containers at L2 |
| 2 | **Boot-window egress block** (`ApplyBootBlockRule`) | restricted/allowlist runs now block egress until real isolation lands |
| 3 | **Pre-boot IPv6 disable** (`DisableIPv6AtBoot`) | closes the IPv6 egress window in restricted/allowlist mode |
| 4 | **`[limits.disk] tmpfs_size`** | `coi run` no longer ENOSPCs on a default `/tmp` where `coi shell` wouldn't |
| 5 | **`[[credentials]]`** seeding | generic host→container secret files (same class as `[[mounts]]`/`[[sockets]]`, which run already honored) |
| 6 | **Git identity** (`SetupGitIdentity` + `useConfigOnly` guard; fail-closed `git.readonly` lock) | `coi run -- git commit` gets the right identity; `git.readonly` is now enforced |

## Where each is applied

- **Pre-boot** (run's existing pre-start hook, which also re-runs on the ephemeral isolation-fallback recreate): NIC security, tmpfs size, pre-boot IPv6 disable (non-open, paired with the already-present `ConfigureNetworkdIPv4Only`).
- **Post-start**: boot block right after first boot (non-open only, fail-closed — removed by the apply-network phase's `SetupForContainer`, same lifecycle as shell); credentials + git identity after `WaitForReady`.
- **Trust**: `[[credentials]]` now flows through `gateRunForwarding` (the credential slot added to `FilterTrusted`), so untrusted project-config credentials are dropped exactly like mounts/sockets/ports.
- Applied on **fresh launches**; a reused persistent container already carries them from its first launch.
- Boot block is deliberately **non-open only**: run's open-mode network path short-circuits and never removes a boot block, so installing one there would strand the container's network.

## End-to-end tests (`tests/run/`, core-build CI group)

One per fix, modeled on the existing `test_run_network_isolation.py` (the #373 precedent for this exact class):
- `run_nic_security.py` — eth0 filtering/port-isolation keys via host `incus config device get` on a persistent run.
- `run_tmpfs_size.py` — `/tmp` device `size` == configured.
- `run_ipv6_disabled_restricted.py` — `disable_ipv6 == 1` inside a restricted run.
- `run_boot_block_restricted_egress.py` — restricted run still reaches `example.com` (boot block installed **and** cleanly removed, not stranded).
- `run_credentials.py` — ad-hoc `[[credentials]]` file present in the container (trusted-scope config).
- `run_git_identity.py` — `user.name`/`user.email` seeded + `user.useConfigOnly == true` (trusted-scope config).

## Verification

`go build ./...`, `go vet`, `gofmt`, `go test ./internal/cli/... ./internal/session/...`, and `ruff check`/`format` all pass.

Depends conceptually on #726/#727 but is independent code (touches `phases_run.go`/`run.go`/`setup_credentials.go` + `tests/run/`; no overlap with #727's `setup.go`/`phases_shell.go`).


mensfeld:
Sorry for the noise.

PR #730, opened 2026-08-26
fix(shell): honor storage_pool config in coi shell

### LINKED ISSUE #726: coi shell ignores [container] storage_pool — only coi run/coi build honor it

technicalpickles:
## Summary

- Fixes #726: `coi shell` silently ignored `[container] storage_pool`, always launching on whichever pool the Incus `default` profile points at. `coi run`/`coi build` already honored it via `launchOrReuseContainer`; this brings `shell` to parity.
- Adds `StoragePool` to `session.SetupOptions` and threads it into the `incus init -s <pool>` call in `session.Setup()`.
- Wires `a.cfg.Container.StoragePool` into `SetupOptions{}` in `phases_shell.go`'s `configureSessionPhase`.
- Validates the pool up front in `validateEnvPhase` (same `container.ValidateStoragePool` check `run` already does), so a bad pool fails fast with an actionable `incus storage create` error instead of failing deep inside Incus mid-launch.

## Test plan

- [x] New Go integration test `TestSetup_HonorsStoragePool` (`internal/session/storage_pool_integration_test.go`): launches a real container via `session.Setup()` with `StoragePool` set and confirms it lands on that pool (checked via `incus list -c b`). Confirmed RED on unpatched master, GREEN with the fix.
- [x] New Python CLI test `tests/cli/test_storage_pool_validation_shell_missing.py`: mirrors the existing `run` validation test but for `shell`, confirming it fails in <0.1s for a nonexistent pool (proving validation fires before any container work). Confirmed RED on unpatched master, GREEN with the fix.
- [x] Manual end-to-end: `coi shell --debug` with `storage_pool = "btrfs-pool"` in a throwaway workspace launched the container directly on `btrfs-pool` (verified via `incus list`), not `default`.
- [x] Regression: `coi run` and `coi build` still work against the same config.
- [x] `go build ./...` / `go vet ./...` clean.
- [x] Full `internal/session` Go suite passes, aside from two pre-existing failures confirmed to reproduce identically on unmodified master (a codex-config trailing-newline seeding bug, and a UID-namespace-isolation race tied to a specific host UID) — unrelated to this change.
- [x] `tests/cli/test_storage_pool_validation_*.py` all pass.

https://claude.ai/code/session_01Qr1Ckjh7ybT7FSWjC27Jvh

### DIFF
--- a/internal/cli/phases_shell.go
+++ b/internal/cli/phases_shell.go
@@ -139,6 +139,9 @@ func (a *App) validateEnvPhase(cmd *cobra.Command, s *shellState) session.Phase 
 			if warning := container.CheckKernelVersion(); warning != "" {
 				fmt.Fprintf(os.Stderr, "%s\n", warning)
 			}
+			if err := container.ValidateStoragePool(a.cfg.Container.StoragePool); err != nil {
+				return nil, err
+			}
 			return nil, nil
 		},
 	}
@@ -364,6 +367,7 @@ func (a *App) configureSessionPhase(cmd *cobra.Command, s *shellState) session.P
 				WorkspacePath:         s.absWorkspace,
 				SessionName:           a.sessionName(),
 				Image:                 img,
+				StoragePool:           a.cfg.Container.StoragePool,
 				Persistent:            a.persistent,
 				ResumeFromID:          resumeID,
 				Slot:                  slotNum,
--- a/internal/session/setup.go
+++ b/internal/session/setup.go
@@ -31,7 +31,8 @@ type SetupOptions struct {
 	WorkspacePath         string
 	SessionName           string // [container] session_name: keys the session identity instead of the workspace path when set
 	Image                 string
-	Persistent            bool // Keep container between sessions (don't delete on cleanup)
+	StoragePool           string // [container] storage_pool: Incus storage pool to launch on (empty = Incus default)
+	Persistent            bool   // Keep container between sessions (don't delete on cleanup)
 	ResumeFromID          string
 	Slot                  int
 	MountConfig           *MountConfig      // Multi-mount support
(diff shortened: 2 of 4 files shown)


mensfeld:
oh I tackled it already but I will cherry pick some tests out of it so thank you!

mensfeld:
Thanks a lot for this, @technicalpickles — and for the thorough diagnosis in #726! 🙏

Closing as a duplicate: the same `storage_pool` fix already landed in #727 (merged), which threads `SetupOptions.StoragePool` into `incus init -s <pool>` and adds the up-front `ValidateStoragePool` in the shell path — functionally identical to your change here.

Your two tests were nicer than what #727 shipped, so I didn't want to lose them: I've cherry-picked both (the `TestSetup_HonorsStoragePool` integration test and the shell missing-pool validation test) into #732, co-authored to you. Really appreciate the contribution!

Issue #707, opened 2026-08-17
Provide ability to resolve top data somehow with the container context


mensfeld:
It drives me crayz that i have to resolve pids to coi container references to figure out which of the containers is overloading my machine

PR #737, opened 2026-08-27
fix(top): container view aggregates the whole container's cgroup, not just init.scope

mensfeld:
Correction to the **unreleased** `coi top` feature (#707) — folded into that feature's release notes rather than a separate Fixed entry, since it never shipped broken.

## Symptom

`coi top` (container view) reports a near-constant **~8 MB** of memory for every container, while `coi top <name>` / `coi top --procs` (per-process view) correctly sums to **hundreds of MB** (e.g. a `claude` process alone at ~470 MB). CPU/IO under-counted the same way.

```
coi top:
coi-51a8604a-2  -  0%  3.2 MB  ...        # container view — wrong

coi top --procs (same container):
coi-51a8604a-2  ...  426.8 MB  claude ...
coi-51a8604a-2  ...   64.3 MB  dockerd ...   # processes sum to hundreds of MB
```

## Root cause

`CollectResourceStats` reads `memory.current` / `cpu.stat` / `io.stat` from the raw path `GetCgroupPath` returns. On systemd-based containers, `GetCgroupPath`'s `incus info` fallback resolves to the **init process's own `init.scope` sub-cgroup**, which accounts for only PID 1 — hence the ~8 MB floor.

The **process view was already correct** because `collectProcessesViaHostProc` strips the trailing systemd scope/service segment to reach the container root before matching processes. The container view never did that strip, so the two views disagreed.

## Fix

- Extract the strip into a shared `containerRootCgroupPath` helper and use it in `CollectResourceStats`, so both views resolve the same **container-root** cgroup — whose cgroup v2 counters aggregate the entire process tree.
- Remove the I/O read's one-level parent climb: it existed to escape `init.scope` (which tracks no I/O). Now that the root is read directly, climbing further would reach the grouping cgroup shared by **all** containers and over-count.

## Changelog

No separate Fixed entry — the whole-tree aggregation is noted directly in the unreleased #707 feature bullet.

## Tests

- New unit test for `containerRootCgroupPath` (init.scope/service stripped; already-root paths unchanged).
- `go build`, `go test ./internal/monitor/ ./internal/cli/`, `gofumpt -l`, `golangci-lint run` all clean locally.


mensfeld:
Added a container-backed regression test (`test_top_container_memory_aggregates_whole_tree` in `tests/top/test_top.py`, in the CI misc-commands matrix): launches a real container and asserts the container-view `memory_mb` (1) covers the largest per-process RSS and (2) comfortably exceeds the init process's own RSS — both impossible under the old init.scope-only read.

PR #778, opened 2026-09-06
fix(image): look up published image in the configured Incus project

### LINKED ISSUE #777: Image build fails in non-default Incus project: getImageFingerprint hardcodes --project default (v0.11.2)
### Summary

On a host where COI's configured Incus project is anything other than `default`, both `coi build` and `coi build --profile <name>` fail after the image publish step. `incus publish` correctly lands the image in the configured project, but the immediately following fingerprint lookup queries the hardcoded project `default` and errors.

### Version

code-on-incus v0.11.2 (also present on master as of this report)

### Environment

- Incus (LTS) with COI configured via `[incus] project = "development"` in config.toml
- `incus image list --project development` shows the freshly published timestamped image with its alias; project `default` contains no such image
- Exit 1 from `coi build`; no bare `coi-default` rolling alias and no versioned main alias are created

### Observed behavior

`coi build` publishes successfully, then fails:

`image not found: coi-default-20260906-060222`

### Root cause

`internal/image/builder.go` `getImageFingerprint` overrides the project for its lookup instead of using the configured project:

`internal/image/builder.go:651` (v0.11.2; identical on master):

```go
output, err := container.IncusOutput("image", "list", alias, "--project", "default", "--format=json")
```

Every other Incus call goes through `buildIncusCommand` (internal/container/commands.go), which injects `container.IncusProject` (set from `[incus] project` via `container.Configure`). `getImageFingerprint` is the only call site that overrides it. Upstream default `[incus] project` is `"default"` (internal/container/commands.go:22), which is why this does not reproduce in the default-project setup.

Both callers are affected:

- `createImage` (builder.go:600) — fails right after `incus publish` succeeds
- `updateAlias` (builder.go:631) — fails before the bare rolling alias (e.g. `coi-default`) is created or repointed

### Suggested fix

Use the configured project in the lookup, consistent with every other call:

```go
output, err := container.IncusOutput("image", "list", alias, "--project", container.IncusProject, "--format=json")
```

### Notes

- The build container and cleanup run in the configured project; only this lookup escapes it.
- A workaround on affected hosts is currently impossible without patching COI (no config surfaces the project for this call), so non-default-project users cannot complete any image build.

marshalfevzi:
Fixes #777

### Problem

On any host where `[incus] project` is not `default`, `coi build` (and `coi build --profile <name>`) fails right after a **successful** `incus publish`:

```
image not found: coi-default-20260906-060222
```

The published image lands in the configured project (via the project-aware wrapper layer), but `getImageFingerprint` then looks it up in the hardcoded project `default` and errors. Both callers are affected:

- `createImage` (internal/image/builder.go:600) — fails immediately after publish; the timestamped image exists but the build reports failure (exit 1)
- `updateAlias` (internal/image/builder.go:631) — fails before the bare rolling alias (e.g. `coi-default`) is created or repointed, so downstream consumers never get the expected alias

### Root cause

`getImageFingerprint` was the only call site overriding the project-aware command construction. Every other Incus invocation routes through `buildIncusCommand` (internal/container/commands.go), which injects `container.IncusProject` — set from `[incus] project` via `container.Configure` (also honored by the explicit `--project` plumbing in internal/cli/audit.go:211). Upstream's default `[incus] project` is `"default"` (internal/container/commands.go:22), so the defect is invisible in default-project setups.

### Fix

Use the configured project, consistent with every other call site:

```go
output, err := container.IncusOutput("image", "list", alias, "--project", container.IncusProject, "--format=json")
```

One line. No behavior change for default-project users (the configured project already defaults to `"default"`).

### Release impact

The defect ships in v0.11.2 (and earlier non-default-project builds). Non-default-project users cannot complete any image build today; there is no config workaround since this call bypasses the configured project entirely. I'd suggest it qualifies for a patch release.

### Verification

- `go build ./...` could not be run locally by the reporter (no Go toolchain at hand); the change is a same-package identifier swap (`"default"` → `container.IncusProject`), with `container` already imported in this file and `IncusProject` used elsewhere in the codebase.
- Reviewers: please run `go build ./...` and `go test ./internal/image/...` as usual. Happy to adjust if you prefer threading the project explicitly through `Builder` opts instead of the package-level variable.


mensfeld:
Should be merged and released today.

mensfeld (dismissed):

mensfeld:
Thank you.

